PR #2, opened 2019-11-12
CTCTRADERS-112: Make InterchangeControlReference number unique

chris-ianson:
SequentialInterchangeControlReferenceIdService is used to generate a unique InterchangeControlReference. Run the it tests with sbt it:test

Also added a new datetime argument to the buildXml method (we don't use the InterchangeControlReference model to get the datetime now, makes more sense to pass it into the buildXml() method.


app/models/messages/request/InterchangeControlReference.scala
```diff
@@ -16,13 +16,9 @@
 
 package models.messages.request
 
-import java.time.LocalDateTime
-
-import utils.Format
-
-case class InterchangeControlReference(prefix: String, dateTime: LocalDateTime) {
+case class InterchangeControlReference(prefix: String, date: String, index: Int) {
```
amaalali: Do you want the prefix to be hard-coded here for now?

PR #3, opened 2019-11-13
CTCTRADERS-108: XSD validation for ArrivalNotification XML 

MatthewAlexanderBeck:


app/utils/ValidateXML.scala
```diff
@@ -0,0 +1,83 @@
...
+  private val schemaLang = javax.xml.XMLConstants.W3C_XML_SCHEMA_NS_URI
+
+  private def saxParser(schema: Schema): SAXParser = {
+    val saxParser: SAXParserFactory = javax.xml.parsers.SAXParserFactory.newInstance()
+    saxParser.setNamespaceAware(true)
+    saxParser.setSchema(schema)
+    saxParser.newSAXParser()
+  }
+
+  def validate(xml: String, fileKey: String): Try[String] = {
```
chris-ianson: Maybe the fileKey could be an XSDFile trait which contains the full resource path. You could then have something like _ArrivalNotificationXSD_ which extends of this trait.

app/models/messages/XSDFile.scala
```diff
@@ -0,0 +1,26 @@
+/*
```
chris-ianson: Do you think the XSDFile is in the right place? It's within messages directory but isn't a message.

app/utils/ValidateXML.scala
```diff
@@ -0,0 +1,83 @@
+/*
```
chris-ianson: Could the ValidateXML be made into a service?

test/models/messages/ArrivalNotificationSpec.scala
```diff
@@ -432,5 +431,97 @@ class ArrivalNotificationSpec extends FreeSpec with MustMatchers
           Json.toJson(notification: ArrivalNotification) mustEqual json
       }
     }
+
+    "must convert to valid xml" in {
```
chris-ianson: If we move the xsd into a service we could create another spec and move all validation checks into this

app/models/messages/XSDFile.scala
MatthewAlexanderBeck: Agreed, moved outside of messages directory.

PR #4, opened 2019-11-14
CTCTRADERS-115: Receive arrival notification

chris-ianson:
Created initial controller interface to accept frontend submissions. We know what's left to do on this branch but will raise smaller PRs so they're easier to review.

NOTE: Property testing isn't within this branch, when the models have been updated to generate more realistic test data we'll implement this (the current generators threw 413 errors when using property testing)

### DIFF
--- a/app/services/SubmissionService.scala
+++ b/app/services/SubmissionService.scala
@@ -14,18 +14,17 @@
  * limitations under the License.
  */
 
-package config
+package services
 
-import com.google.inject.AbstractModule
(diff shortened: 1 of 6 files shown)


app/services/SubmissionService.scala
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package services
+
+import com.google.inject.Singleton
+import models.messages.ArrivalNotification
+
+@Singleton
```
MatthewAlexanderBeck: Can this be removed as we add it to the Module as an eager singleton?

PR #5, opened 2019-11-15
CTCTRADERS-115: create presist and delete to mongo functionality for arrival notifications

MatthewAlexanderBeck:


it/services/ArrivalNotificationServiceSpec.scala
```diff
@@ -0,0 +1,138 @@
...
+
+  "ArrivalNotificationService" - {
+
+    "must persist NormalNotification within mongoDB" in {
+
+      forAll(arbitrary[NormalNotification]) {
+
+        normalNotification =>
+
+          database.flatMap(_.drop()).futureValue
```
chris-ianson: I know we use this in a couple of places. Could we mix in BeforeAndAfter trait and use the beforeEach method to drop the database.

app/services/ArrivalNotificationService.scala
```diff
@@ -0,0 +1,60 @@
+/*
```
chris-ianson: Do you think we should move this into an ArrivalNotificationRepository? I'm just thinking we could have a separate service which calls the repository and keeps the controller and persistence layers apart. The service could be responsible for how to handle responses from mongo.

PR #12, opened 2019-12-02
CTCTRADERS-170: Handle enroute event incidents 

chris-ianson:


app/models/EnRouteEvent.scala
```diff
@@ -21,7 +21,7 @@ import play.api.libs.json._
 final case class EnRouteEvent(
   place: String,
   countryCode: String,
-  alreadyInNcts: Boolean,
+  alreadyInNcts: Int,
```
chris-ianson: We might have to check what the frontend is sending us here. Could be that the frontend needs updating. Also, Boolean type does indicate it can only be one of two values. Do we think an Int indicates the intent?

app/services/XmlBuilderService.scala
```diff
@@ -126,7 +126,9 @@ class XmlBuilderService {
             <PlaTEV10>{event.place}</PlaTEV10>
             <PlaTEV10LNG>{arrivalNotificationRequest.header.languageCode}</PlaTEV10LNG>
             <CouTEV13>{event.countryCode}</CouTEV13>
-            <CTLCTL>{event.alreadyInNcts}</CTLCTL>
+            <CTLCTL>
```
chris-ianson: Good shout on this one!

app/models/EnRouteEvent.scala
MatthewAlexanderBeck: Yeah, think you might be right here. This is maybe a question we can pass back to the BA's as it may be the case that the Int's are going to be either 1 or 0 depending on if true or false

PR #13, opened 2019-12-04
CTCTRADERS-169

kenneth-rayner:


app/controllers/ArrivalNotificationController.scala
```diff
@@ -77,23 +79,35 @@ class ArrivalNotificationController @Inject()(
 
                         }
                         .recover {
-                          case _ => InternalServerError
+                          case _ => {
+                            InternalServerError(Json.toJson(ErrorResponseBuilder.failedSavingArrivalNotification))
+                              .as("application/json")
+                          }
```
MatthewAlexanderBeck: We could pass the exception into the failSavingArrivalNotification to help the frontend indicate is either the mongo has failed or the POST to the hods.

Not necessary but it could be helpful for troubleshooting.

app/controllers/ArrivalNotificationController.scala
```diff
@@ -77,23 +79,35 @@ class ArrivalNotificationController @Inject()(
...
                         }
                     }
                     case Left(FailedToValidateXml) =>
-                      Future.successful(BadRequest)
+                      Future.successful(
+                        BadRequest(Json.toJson(ErrorResponseBuilder.failedXmlValidation))
+                          .as("application/json"))
                   }
```
MatthewAlexanderBeck: Do we have any plan's to pass the exact problem from the validation error (and all other errors) to the frontend, will this be done in this ticket or should we complete this at a later date. 

Again not necessary but it would help the frontend  and out logging of any problems with our xml and data modelling.

PR #18, opened 2019-12-06
Update fmt config file  and related changes

ashwinirudrappa:


app/services/XmlBuilderService.scala
```diff
@@ -53,99 +52,105 @@ class XmlBuilderService {
...
+        </PlaTEV10> <PlaTEV10LNG>
+        {arrivalNotificationRequest.header.languageCode}
+      </PlaTEV10LNG> <CouTEV13>
+        {event.countryCode}
+      </CouTEV13> <CTLCTL>
+        <AlrInNCTCTL29>
+          {if (event.alreadyInNcts) 1 else 0}
+        </AlrInNCTCTL29>
+      </CTLCTL>{buildIncident(event.eventDetails)(arrivalNotificationRequest)}
+      </ENROUEVETEV>
```
MatthewAlexanderBeck: Would it be possible to revert this back please, intellij formats makes a mess of xml. Would it be possible to do this in test as well please.

PR #20, opened 2019-12-11
Character encoding for Xml submission

MatthewAlexanderBeck:


app/services/XmlBuilderService.scala
```diff
@@ -204,21 +214,31 @@ object XmlBuilderService {
...
+    }
+    case _ => NodeSeq.Empty
+  }
+
+  private def buildOptionalDate(value: Option[LocalDate], elementTag: String): NodeSeq = value match {
     case Some(result: LocalDate) => loadString(s"<$elementTag>${Format.dateFormatted(result)}</$elementTag>")
     case _                       => NodeSeq.Empty
   }
 
+  private def buildAndEncodeElem[A](value: A, elementTag: String): NodeSeq = value match {
```
chris-ianson: I like how you're using the type to determine if encoding is needed. Do you think we should create a new type for _Language Code_ We know it's only ever EN and if we have a custom type we can also exclude it from any encoding. Thoughts?

test/services/XmlBuilderServiceSpec.scala
```diff
@@ -191,19 +336,34 @@ class XmlBuilderServiceSpec
...
+    }
+    case result: Boolean       => loadString(s"<$elementTag>${if (result) 1 else 0}</$elementTag>")
+    case result: LocalDate     => loadString(s"<$elementTag>${Format.dateFormatted(result)}</$elementTag>")
+    case result: LocalDateTime => loadString(s"<$elementTag>${Format.dateFormatted(result)}</$elementTag>")
+    case _                     => NodeSeq.Empty
   }
 
   private def buildIncidentFlag(hasIncidentInformation: Boolean): NodeSeq = hasIncidentInformation match {
-    case false => <IncFlaINC3>1</IncFlaINC3>
+    case false => <IncFlaINC3>{"1"}</IncFlaINC3>
```
chris-ianson: Why is this wrapped in braces?

test/generators/MessageGenerators.scala
```diff
@@ -32,12 +32,6 @@ import org.scalacheck.Gen
 
 trait MessageGenerators extends ModelGenerators {
 
-  implicit lazy val validXmlString: Arbitrary[String] =
```
chris-ianson: So the generators should test that characters are being properly encoded 👍

PR #25, opened 2020-01-20
CTCTRADERS-142: Convert seals to xml

ashwinirudrappa:


app/services/XmlBuilderService.scala
```diff
@@ -132,15 +132,36 @@ class XmlBuilderService {
...
+  private def buildSeals(seals: Option[Seq[Seal]]): NodeSeq =
+    seals.fold(NodeSeq.Empty) {
+      seals =>
+        val sealsXml = seals.map {
+          seal =>
+            <SEAIDSI1>
+            <SeaIdeSI11>
+              {seal.numberOrMark}
+            </SeaIdeSI11>
+            <SeaIdeSI11LNG>GB</SeaIdeSI11LNG>
```
garethgooch: this should be EN not GB, check with CL012 for language codes, does this need hard coded or would this be DE is applied in Germany, I think its OK but needs to be checked.

test/services/XmlValidationServiceSpec.scala
```diff
@@ -34,27 +36,12 @@ class XmlValidationServiceSpec extends SpecBase {
...
-        val xml = buildXml(withEnrouteEvent = true, withContainerTranshipment = true)
-
-        xmlValidationService.validate(xml, ArrivalNotificationXSD) mustBe a[Right[_, _]]
-      }
-
-      "with an enroute event and vehicular transhipment" in {
-        val xml = buildXml(withEnrouteEvent = true, withVehicularTranshipment = true)
-
-        xmlValidationService.validate(xml, ArrivalNotificationXSD) mustBe a[Right[_, _]]
+        forAll(arbitrary[Boolean], arbitrary[Boolean], arbitrary[Boolean], arbitrary[Boolean]) {
```
amaalali: The`withIncident `  is not being used, which means we're not testing this branch of logic, I think we either break this out into two tests: 

1. "with an enroute event without incident"
2. "with an enroute event with incident"

or make the test with arbitrary `withIncident` value work

app/services/XmlBuilderService.scala
```diff
@@ -132,15 +132,36 @@ class XmlBuilderService {
...
+               buildIncident(event.eventDetails, event.seals)(arrivalNotificationRequest)
             }
           </ENROUEVETEV>
       }
   }
 
-  private def buildIncident(event: EventDetails)(implicit arrivalNotificationRequest: ArrivalNotificationRequest): NodeSeq = event match {
-    case incident: Incident =>
-      <INCINC> {
+  private def buildSeals(seals: Option[Seq[Seal]]): NodeSeq =
```
amaalali: I think we need to invert the control here, it should be the concern of the `buildSeals` function to construct a valid `NodeSeq` given a `Seq[Seal]`. In its current form, it is also introspecting on its parameter (`seals: Option[Seq[Seal]] `) and checking if it needs to default, which feels like a concern of the orchestrator which is the `buildIncident` method.

If we do this though, we would need to move the defaulting of the `NodeSeq.Empty` to the place where this method is called

app/services/XmlBuilderService.scala
```diff
@@ -151,10 +172,11 @@ class XmlBuilderService {
           buildAndEncodeElem(arrivalNotificationRequest.header.languageCode, "EndPlaINC10LNG") ++
           buildOptionalElem(incident.endorsement.country, "EndCouINC12")
         }
-      </INCINC>
+      </INCINC> ++ buildSeals(seals)
```
amaalali: This is an example of what we would need to change in the previous comment

test/services/XmlValidationServiceSpec.scala
ashwinirudrappa: @amaalali Good catch, we will look into it

PR #27, opened 2020-01-24
Added Constants for generators and language codes

MatthewAlexanderBeck:


app/models/request/Header.scala
```diff
@@ -22,8 +22,15 @@ case class Header(
...
 
-sealed trait HeaderConstants {
-  val languageCode: LanguageCode = LanguageCodeEnglish
+object Header {
+
+  object Constants {
+    val languageCode: LanguageCode     = LanguageCodeEnglish
+    val customsSubPlaceLength          = 17
+    val arrivalNotificationPlaceLength = 35
+    val simplifiedProcedureFlag0       = "0"
```
chris-ianson: You think we should update simplifiedProcedureFlag to a custom type?

MatthewAlexanderBeck: Agreed, will apply the change now.

PR #45, opened 2020-03-10
CTCTRADERS-319: Added eori auth to ArrivalNotificationController

MatthewAlexanderBeck:


app/controllers/actions/IdentifierAction.scala
```diff
@@ -0,0 +1,65 @@
...
+
+class AuthenticatedIdentifierAction @Inject()(
+  override val authConnector: AuthConnector,
+  config: AppConfig,
+  val parser: BodyParsers.Default
+)(implicit val executionContext: ExecutionContext)
+    extends IdentifierAction
+    with AuthorisedFunctions {
+
+  private val enrolmentIdentifierKey: String = "VATRegNoTURN"
```
chris-ianson: Would this be better as a config value? (we know this is likely to change when we introduce ROSM work)

PR #46, opened 2020-03-10
Decouple XML from controller

amaalali:


app/services/SubmissionModelService.scala
```diff
@@ -43,15 +51,13 @@ class SubmissionModelService @Inject()(appConfig: AppConfig) {
         Left(FailedToConvertModel)
     }
 
-  private def buildMeta(messageSender: MessageSender, interchangeControlReference: InterchangeControlReference): Meta =
-    Meta(messageSender = messageSender, interchangeControlReference = interchangeControlReference)
-
   private def buildHeader(arrivalNotification: NormalNotificationMessage, procedureTypeFlag: ProcedureTypeFlag): Header =
```
garethgooch: Should this be pulled out like Meta/buildMeta has?

app/utils/Format.scala
```diff
@@ -27,8 +28,9 @@ object Format {
   def dateFormatted(date: LocalDate): String     = date.format(dateFormatter)
   def dateFormatted(date: LocalDateTime): String = date.format(dateFormatter)
 
-  val timeFormatter: DateTimeFormatter           = DateTimeFormatter.ofPattern("HHMM")
+  val timeFormatter: DateTimeFormatter           = DateTimeFormatter.ofPattern("HHmm")
   def timeFormatted(date: LocalDateTime): String = date.format(timeFormatter)
+  def timeFormatted(date: LocalTime): String     = date.format(timeFormatter)
```
garethgooch: should this be an input of time rather than date?

app/services/SubmissionModelService.scala
garethgooch: should this be broken out like buildMeta has been?

test/models/request/ArrivalNotificationRequestSpec.scala
```diff
@@ -45,17 +46,20 @@ class ArrivalNotificationRequestSpec
     with ScalaCheckDrivenPropertyChecks
     with OptionValues {
 
-  private val dateTime: LocalDateTime = LocalDateTime.now()
-  private val dateOfPreparation       = Format.dateFormatted(dateTime)
-  private val timeOfPreparation       = Format.timeFormatted(dateTime)
+  private val dateOfPreparation          = LocalDate.now()
+  private val dateOfPreparationFormatted = Format.dateFormatted(dateOfPreparation)
```
garethgooch: This would still be clear as a single line, private val dateOfPreperation = Format.dateFormatted(LocalDate.now())

test/models/request/ArrivalNotificationRequestSpec.scala
```diff
@@ -45,17 +46,20 @@ class ArrivalNotificationRequestSpec
...
+  private val timeOfPreparation          = LocalTime.now()
+  private val timeOfPreparationFormatted = Format.timeFormatted(timeOfPreparation)
```
garethgooch: ^^ Ditto ^^

app/utils/Format.scala
amaalali: Do you mean the name of the parameter? In which case yes

app/services/SubmissionModelService.scala
amaalali: Not sure, I don't think it matters really, I don't want to create unnecessary changes that are more preferences than anything else

test/models/request/ArrivalNotificationRequestSpec.scala
amaalali: `dateOfPreparation ` is used below to build the expect statement

test/models/request/ArrivalNotificationRequestSpec.scala
amaalali: `timeOfPreparation` is used below to build the expect statement

PR #61, opened 2020-03-20
Submit to upstream and reject non-007 messages

amaalali:


app/controllers/MovementsController.scala
```diff
@@ -64,7 +74,6 @@ class MovementsController @Inject()(
                 InternalServerError
```
chris-ianson: We'll need some logging to understand what the error was because it may have been on the insert or post

app/services/ArrivalMovementService.scala
```diff
@@ -60,22 +60,28 @@ class ArrivalMovementService @Inject()(arrivalIdRepository: ArrivalIdRepository)
...
       .map(Try(_))
       .map(_.map(LocalTime.parse(_, Format.timeFormatter)))
       .map(_.toOption) // TODO: We are not propagating this failure back, do we need to do this?
 
   val mrnR: Reader[NodeSeq, Option[MovementReferenceNumber]] =
-    Reader[NodeSeq, NodeSeq](_ \ "CC007A" \ "HEAHEA" \ "DocNumHEA5")
+    Reader[NodeSeq, NodeSeq](_ \ "HEAHEA" \ "DocNumHEA5")
       .map(_.text)
       .map {
         case mrnString if !mrnString.isEmpty => Some(MovementReferenceNumber(mrnString))
```
chris-ianson:
```suggestion
        case mrnString if mrnString.nonEmpty => Some(MovementReferenceNumber(mrnString))
```

test/controllers/MovementsControllerSpec.scala
```diff
@@ -196,6 +242,47 @@ class MovementsControllerSpec extends SpecBase with ScalaCheckPropertyChecks wit
           verify(mockArrivalMovementRepository, times(0)).insert(any())
         }
       }
+
+      "must return InternalServerError if sending the message upstream fails" in {
```
chris-ianson: It would be good to know if the error is within our service or EIS (i know you'd mentioned something about adding specific logging for this). Just something to keep in mind

PR #62, opened 2020-03-20
Set the State of an arrival notification

TheNickWilson:


app/controllers/MovementsController.scala
```diff
@@ -60,11 +62,23 @@ class MovementsController @Inject()(
...
-                          // TODO: This needs to be replaced url to arrival movement resource, for which we need an Arrival Movement number
-                            .withHeaders("Location" -> arrival.arrivalId.index.toString)
+                          val newState = arrival.state.transition(SubmissionResult.Success)
+                          arrivalMovementRepository.setState(arrival.arrivalId, newState).map {
+                            _ =>
+                              Accepted("Message accepted")
+                              // TODO: This needs to be replaced url to arrival movement resource, for which we need an Arrival Movement number
+                                .withHeaders("Location" -> arrival.arrivalId.index.toString)
+                          }
+                      }
```
amaalali: We should check that the status is 202 here or else it should be a rejected

amaalali: I've checked and we're using the Http reads that throw errors so this is valid. So all good 👍

PR #65, opened 2020-03-23
CTCTRADERS-327:Change method parse.Xml to parse.tolerantXml to read x…

ashwinirudrappa:
…ml body as string


amaalali:
After investigating, HTTP verbs can only send data encoded as JSON. So this change is now the only way forward :(

PR #69, opened 2020-03-24
Changes to how we persist an Arrival 

amaalali:
- Change ArrivalId reads and writes to JsNumber
- Change ArrivalId Json reads and writes to write the arrival field as the "_id" primary index
- Refactor ArrivalIdRepository to read ArrivalId directly

### DIFF
--- a/app/models/Arrival.scala
+++ b/app/models/Arrival.scala
@@ -17,13 +17,29 @@
 package models
 
 import models.request.ArrivalId
-import play.api.libs.json.Json
-import play.api.libs.json.OFormat
+import play.api.libs.json._
+import play.api.libs.functional.syntax._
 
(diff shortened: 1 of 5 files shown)


app/models/Arrival.scala
```diff
@@ -17,13 +17,29 @@
...
+    )(Arrival.apply _)
+
+  implicit val writesArrival: OWrites[Arrival] =
+    (
+      (__ \ "_id").write[ArrivalId] and
+        (__ \ "movementReferenceNumber").write[String] and
+        (__ \ "eoriNumber").write[String] and
+        (__ \ "state").write[State] and
+        (__ \ "messages").write[Seq[MovementMessage]]
+    )(unlift(Arrival.unapply))
```
MatthewAlexanderBeck: Could you not use OFormat for this?

amaalali: We're writing "arrivalId" as "_id", so it wouldn't work with the Json.format

PR #73, opened 2020-03-27
CTCTRADERS-351:Create endpoint to return details for all arrival movements for a use

ashwinirudrappa:


test/controllers/MovementsControllerSpec.scala
```diff
@@ -339,5 +340,53 @@ class MovementsControllerSpec extends SpecBase with ScalaCheckPropertyChecks wit
...
+
+              status(result) mustEqual OK
+              contentAsJson(result) mustEqual Json.toJson(arrivals)
+
+              reset(mockArrivalMovementRepository)
+          }
+        }
+      }
+
+      "must return an INTERNAL_SERVER_ERROR on fail" in {
```
amaalali:
```suggestion
      "must return an INTERNAL_SERVER_ERROR when we cannot retrieve the Arrival Movements" in {
```

test/controllers/MovementsControllerSpec.scala
```diff
@@ -339,5 +340,53 @@ class MovementsControllerSpec extends SpecBase with ScalaCheckPropertyChecks wit
       }
     }
 
+    "getArrivals" - {
+
+      "must return Ok and retrieve arrivals" in {
```
amaalali:
```suggestion
      "must return Ok with the retrieved arrivals" in {
```

test/services/ArrivalMovementServiceSpec.scala
```diff
@@ -36,14 +38,16 @@ import scala.concurrent.Future
 
 class ArrivalMovementServiceSpec extends SpecBase with IntegrationPatience {
 
+  val dateOfPrep          = LocalDate.now()
+  val timeOfPrep          = LocalTime.of(1, 1)
+  private val arrivalMeta = ArrivalMeta(ArrivalDateTime(dateOfPrep, timeOfPrep), ArrivalDateTime(dateOfPrep, timeOfPrep))
```
amaalali: Can we move these to the only test that it's being used in:     "creates an arrival movement with an internal ref number and a mrn, date and time of creation from the message submitted"

PR #75, opened 2020-03-31
CTCTRADERS-338: Added XSD validation for GoodsReleased

MatthewAlexanderBeck:


app/services/XmlValidationService.scala
```diff
@@ -0,0 +1,74 @@
...
+
+      xmlResponse.parser.parse(new InputSource(new StringReader(xml)), new CustomParseHandler())
+
+      Success("successfully parsed xml")
+
+    } catch {
+      case e: Throwable =>
+        logger.warn(e.getMessage)
+        Failure(e)
+    }
```
amaalali, line 72: I'm not sure the logging makes sense here. The XmlValidationService doesn't have context about the validation being done here, instead, the caller should do the logging because it will have the context. The signature of the method already allows for this inversion of control as it's returning a Try with the error that the caller can interpret.

Also, is it possible to replace the `try ... catch` with Try's apply method which does this for us but provides a nice way to interact with it?

conf/application.conf
```diff
@@ -141,3 +141,9 @@ microservice {
   }
 
 }
+
+xsd {
+    goodsReleased = "/xsd/CC025A.xsd"
+}
```
amaalali, line 147: Anything that's in here is overwritable, but we don't want these values to be overwritten in configs in other environments, should we move it out into code?

test/controllers/GoodsReleasedControllerSpec.scala
```diff
@@ -29,63 +29,76 @@ import models.State
...
+      <DatOfPreMES9>{Format.dateFormatted(dateOfPrep)}</DatOfPreMES9>
+      <TimOfPreMES10>{Format.timeFormatted(timeOfPrep)}</TimOfPreMES10>
+    </CC025A>
+
+  override def beforeEach: Unit = {
+    super.beforeEach()
+    reset(mockArrivalMovementRepository)
+    reset(mockLockRepository)
+    reset(mockXmlValidationService)
+  }
```
amaalali, line 80: I'm not sure I see the benefit of doing this? All we've done is create global mutable state that can leak between tests and we've been forced to bring in test hooks to clean up that global mutable data.

test/base/SpecBase.scala
```diff
@@ -41,4 +43,8 @@ trait SpecBase extends FreeSpec with MustMatchers with MockitoSugar with ScalaFu
         bind[AuthenticateActionProvider].to[FakeAuthenticateActionProvider]
       )
 
+  def injector: Injector = baseApplicationBuilder.injector
+
+  def appConfig = injector.instanceOf[AppConfig]
```
amaalali, line 48: Can we remove this? This is Injector and AppConfig was previously removed because in many tests, there were overriding module bindings but the tests used this injector which had different bindings from the tests. Instead in the tests where this is being used, they can just use the app's injector for that test.

test/services/XmlValidationServiceSpec.scala
```diff
@@ -0,0 +1,132 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package services
+
+import base.SpecBase
+
+class XmlValidationServiceSpec extends SpecBase {
```
amaalali, line 21: This service is testing against a specific XSD, which means that anyone looking/maintaining at this test will need to look at that XSD, understand it and construct valid XML sample for that XSD. Instead, should we have a simple XSD that we can easily construct XML data for and which is purposely designed to be simple?

test/services/XmlValidationServiceSpec.scala
```diff
@@ -0,0 +1,132 @@
...
+      "must return a failure when given an xml with invalid fields" in {
+
+        val invalidXml = {
+          <CC025A>
+            <SynIdeMES1>UNOC</SynIdeMES1>
+            <SynVerNumMES2>3</SynVerNumMES2>
+            <MesSenMES3>NTA.GB</MesSenMES3>
+            <MesRecMES6>SYST17B-NCTS_EU_EXIT</MesRecMES6>
+            <DatOfPreMES9>20191004</DatOfPreMES9>
+            <TimOfPreMES10>Invalid field</TimOfPreMES10>
```
amaalali, line 101: Is this the node that is invalid? It's not obvious from this test what the invalid field is

PR #82, opened 2020-04-16
Established New Database Schema, added POST and PUT endpoints

RobertLaverick:


app/controllers/MovementsController.scala
```diff
@@ -20,87 +20,166 @@ import java.time.OffsetDateTime
...
   authenticate: AuthenticateActionProvider,
+  authenticateForWrite: AuthenticatedGetArrivalForWriteActionProvider,
   messageConnector: MessageConnector
 )(implicit ec: ExecutionContext)
     extends BackendController(cc) {
 
   private val logger = Logger(getClass)
 
-  def createMovement: Action[NodeSeq] = authenticate().async(parse.xml) {
+  def createMovement: Action[NodeSeq] = authenticatedOptionalArrival().async(parse.xml) {
```
chris-ianson, line 69: This is called createMovement but further down there's a method for getArrivals. Maybe consider renaming this (it's doing more than just creating a movement)

chris-ianson, line 69: Noticed we have multiple authentication actions. Does this indicate that we may need multiple controllers rather than just a single one that handles every action

app/controllers/MovementsController.scala
```diff
@@ -20,87 +20,166 @@ import java.time.OffsetDateTime
...
     implicit request =>
-      arrivalMovementService.makeArrivalMovement(request.eoriNumber)(request.body) match {
+      request.arrival match {
+        case Some(arrival) => update_internal(arrival, request.body)
```
chris-ianson, line 72: Naming convention isn't followed here. What does this method do?

app/controllers/MovementsController.scala
```diff
@@ -20,87 +20,166 @@ import java.time.OffsetDateTime
...
-                            }
-                      }
-                      .recoverWith {
-                        case error: Exception =>
-                          logger.error(s"Call to EIS failed with the following Exception: ${error.getMessage}")
-                          val newState = arrival.state.transition(SubmissionResult.Failure)
-                          arrivalMovementRepository.setState(arrival.arrivalId, newState).map {
+          arrivalMovementService.makeArrivalMovement(request.eoriNumber)(request.body) match {
+            case None =>
+              Future.successful(BadRequest("Invalid data: missing either DatOfPreMES9, TimOfPreMES10 or DocNumHEA5"))
```
chris-ianson, line 76: Maybe something for future consideration, but is the API going to offer custom error codes - probably not if we're not returning JSON to the consumer of the API but worth thinking about)

app/controllers/MovementsController.scala
```diff
@@ -20,87 +20,166 @@ import java.time.OffsetDateTime
...
+            case Some(x) =>
```
chris-ianson, line 77: Whats x?

app/controllers/MovementsController.scala
```diff
@@ -20,87 +20,166 @@ import java.time.OffsetDateTime
...
+                                    _ <- arrivalMovementRepository.setMessageState(arrival.arrivalId,
+                                                                                   arrival.messages.length - 1,
+                                                                                   state.transition(SubmissionResult.Success))
+                                    _ <- arrivalMovementRepository.setState(arrival.arrivalId, arrival.state.transition(ArrivalSubmitted))
+                                  } yield {
+                                    Accepted("Message accepted")
+                                      .withHeaders("Location" -> routes.MovementsController.get(arrival.arrivalId).url)
+                                  }
+                                case None =>
+                                  Future.successful(InternalServerError)
```
chris-ianson, line 103: Consider pulling this out into it's own method so it's clear it's setting the movement and message state

app/controllers/MovementsController.scala
chris-ianson, line 72: On the Some match you call a private method, could the None also call a private method (could make the public method on the controller easier to understand - it either updates a movement or creates a new one)

app/controllers/actions/AuthenticateGetOptionalArrivalForWriteProvider.scala
```diff
@@ -0,0 +1,88 @@
...
+                          lockRepository.unlock(arrival.arrivalId).map {
+                            _ =>
+                              result
+                          }
+                      }
+                      .recoverWith {
+                        case e: Exception =>
+                          lockRepository.unlock(arrival.arrivalId).map {
+                            _ =>
+                              InternalServerError
```
chris-ianson, line 77: Do we need a log here to indicate it couldn't unlock the record?

conf/app.routes
```diff
@@ -3,4 +3,7 @@
 POST                    /movements/arrivals                 controllers.MovementsController.createMovement
 GET                     /movements/arrivals                 controllers.MovementsController.getArrivals
 
+GET                    /movements/arrivals/:arrivalId                 controllers.MovementsController.get(arrivalId: ArrivalId)
+PUT                    /movements/arrivals/:arrivalId                 controllers.MovementsController.updateArrival(arrivalId: ArrivalId)
```
chris-ianson, line 7: This is going to give us back a different arrival each time it's called, is that right? I know POST vs PUT debate can be a bit hairy so I'll tread carefully here ;D I understand having a POST to handle every request to send a message would require createMovement to handle the internals of adding or creating a message. Is that a bad thing? (you've probably all discussed this to death)

app/controllers/MovementsController.scala
```diff
@@ -20,87 +20,166 @@ import java.time.OffsetDateTime
...
-                // TODO: Add logging here so that we can be alerted to these issues.
-                InternalServerError
+          }
+      }
+
+  }
+
+  private def update_internal(arrival: Arrival, body: NodeSeq)(implicit hc: HeaderCarrier) =
+    arrivalMovementService.makeArrivalNotificationMessage(arrival.nextMessageCorrelationId)(body) match {
+      case None => Future.successful(BadRequest("Invalid data: missing either DatOfPreMES9, TimOfPreMES10 or DocNumHEA5"))
```
ashwinirudrappa, line 136: Its worth revisiting this code for some refactor around common code [method update_internal and createMovement]

app/models/ArrivalState.scala
```diff
@@ -0,0 +1,57 @@
...
+
+package models
+
+sealed trait ArrivalState {
+  def transition(messageReceived: MessageReceived): ArrivalState
+}
+
+object ArrivalState extends Enumerable.Implicits {
+  /*
+    007 -> Default state would be ArrivalSubmitted
```
ashwinirudrappa, line 25: Worth removing the commented code we can note the xml messages type some where else like confluence ...

app/models/ArrivalState.scala
```diff
@@ -0,0 +1,57 @@
...
+ * limitations under the License.
+ */
+
+package models
+
+sealed trait ArrivalState {
+  def transition(messageReceived: MessageReceived): ArrivalState
+}
+
+object ArrivalState extends Enumerable.Implicits {
```
ashwinirudrappa, line 23: Why do we need this class, even in transition it is converting one object to other object, wha is the purpose of this class. 
` case MessageReceived.ArrivalSubmitted => ArrivalSubmitted`

app/models/MovementMessage.scala
```diff
@@ -17,26 +17,72 @@
...
   }
 
-  implicit val formatsMovementMessage: OFormat[MovementMessage] =
-    Json.format[MovementMessage]
+  /*
+  implicit lazy val writes: OWrites[MovementMessage] = OWrites {
+    case ns: MovementMessageWithState    => Json.toJsObject(ns)(MovementMessageWithState.formatsMovementMessage)
+    case ws: MovementMessageWithoutState => Json.toJsObject(ws)(MovementMessageWithoutState.formatsMovementMessage)
+  }
+   */
```
ashwinirudrappa, line 64: Commented code

app/controllers/MovementsController.scala
amaalali: I'll look at splitting them out

app/models/MovementMessage.scala
```diff
@@ -17,26 +17,72 @@
...
 import scala.xml.NodeSeq
-import scala.xml.XML
 
-final case class MovementMessage(dateTime: LocalDateTime, messageType: MessageType, message: NodeSeq)
+abstract class MovementMessage(val dateTime: LocalDateTime, val messageCorrelationId: Int) {
 
-object MovementMessage {
+  def getState(): Option[MessageState] =
+    if (this.isInstanceOf[MovementMessageWithState]) {
+      Some(this.asInstanceOf[MovementMessageWithState].state)
```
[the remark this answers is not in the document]
amaalali: This is being removed entirely as it's a bit of a code smell

app/controllers/MovementsController.scala
```diff
@@ -116,4 +195,7 @@ class MovementsController @Inject()(
             InternalServerError(s"Failed with the following error: $e")
         }
   }
+
+  def get(arrivalId: ArrivalId): Action[AnyContent] = defaultActionBuilder(_ => NotImplemented)
```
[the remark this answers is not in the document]
amaalali: We've added this for the `GET /movements/arrival/{arrivalId}` so that it will resolve in the reverse router. This is used in the `Location` header on a successful `POST /movements/arrival`

app/models/ArrivalState.scala
amaalali: Take a look at how this is used in the `GoodsReleasedController`

app/controllers/MovementsController.scala
```diff
@@ -20,87 +20,166 @@ import java.time.OffsetDateTime
...
+                            MessageSender(arrival.arrivalId, arrival.messages.head.messageCorrelationId)
+                          )
+                          .flatMap {
                             _ =>
-                              BadGateway
+                              arrival.messages.last.getState() match {
+                                case Some(state) =>
+                                  for {
+                                    _ <- arrivalMovementRepository.setMessageState(arrival.arrivalId,
+                                                                                   arrival.messages.length - 1,
```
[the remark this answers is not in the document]
amaalali: That's part of the refactoring that I'm working on after this PR, we do want it to be a single atomic write as opposed to 2 writes which don't give us that atomicity

app/controllers/MovementsController.scala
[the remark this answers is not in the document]
amaalali: We've added this for the `GET /movements/arrival/{arrivalId}` so that it will resolve in the reverse router. This is used in the `Location` header on a successful `POST /movements/arrival`

app/controllers/actions/AuthenticateGetOptionalArrivalForWriteProvider.scala
amaalali: The locks have a ttl of 20 seconds so they should be cleaned up in a window of 20 - 120 seconds. Given that, do you think we need that logging in our locking actions when unlocking fails? It could help us understand if we ever get those error cases where we don't unlock

app/controllers/MovementsController.scala
amaalali: I'll be adding a way to do these both in a single write to mongo so that they get the benefit of being atomic

app/controllers/MovementsController.scala
amaalali: I've gone with `post` for now since this method handles the post essentially

app/controllers/actions/AuthenticateGetOptionalArrivalForWriteProvider.scala
chris-ianson: I'd stick it in otherwise you're just getting an internal server error and we're not sure what happened

amaalali: I'll follow this PR with a separate PR that does that

PR #84, opened 2020-04-24
Added endpoint to get a specific message

RobertLaverick:


app/controllers/MovementsController.scala
```diff
@@ -168,4 +172,14 @@ class MovementsController @Inject()(
             InternalServerError(s"Failed with the following error: $e")
         }
   }
+
+  def getMessage(arrivalId: ArrivalId, messageId: Int): Action[AnyContent] = authenticateForRead(arrivalId) {
+    implicit request =>
+      if (request.arrival.messages.isDefinedAt(messageId) && ((request.arrival
+            .messages(messageId)
+            .isInstanceOf[MovementMessageWithState] && request.arrival.messages(messageId).asInstanceOf[MovementMessageWithState].state == SubmissionSucceeded)
+          || request.arrival.messages(messageId).isInstanceOf[MovementMessageWithoutState]))
```
amaalali, line 181: So I'm not really a big fan that we need to do this 🤔

amaalali, line 181: So I'm not really a big fan that we need to do this 🤔

PR #85, opened 2020-04-24
CTCTRADERS-406: Include unloading permission

garethgooch:
Adding in unloading permission message to be stored and update the information


app/controllers/MessageResponseController.scala
```diff
@@ -0,0 +1,86 @@
...
+                  logger.error(s"Failure to parse message")
+                  Future.successful(InternalServerError)
+              }
+            case Failure(e) =>
+              logger.error(s"Failure to validate against XSD. Exception: ${e.getMessage}")
+              Future.successful(BadRequest)
+          }
+        case None =>
+          logger.error(s"Invalid or empty X-Message-Type")
+          Future.successful(InternalServerError)
```
chris-ianson, line 82: Should this be a BadRequest?  invalid or empty X-Message-Type is a failure by the client (EIS/stub) to provide accepted data

test/services/ArrivalMovementServiceSpec.scala
```diff
@@ -129,7 +129,7 @@ class ArrivalMovementServiceSpec extends SpecBase with IntegrationPatience {
       val messageCorrelationId = 1
       val expectedMessage      = MovementMessageWithoutState(LocalDateTime.of(dateOfPrep, timeOfPrep), MessageType.GoodsReleased, movement, messageCorrelationId)
 
-      service.makeGoodsReleasedMessage(messageCorrelationId)(movement).value mustEqual expectedMessage
+      service.makeMessage(messageCorrelationId, MessageType.GoodsReleased)(movement).value mustEqual expectedMessage
```
chris-ianson, line 132: I know the method is _makeMessage_ but do we need some tests for UnloadingPermission (similar to the what's in the _makeGoodsReleasedMessage_ block

app/controllers/MessageResponseController.scala
chris-ianson: If X-Message-Type is passed in but not handled, log it out here (thanks Gareth!)

app/models/MessageResponse.scala
```diff
@@ -0,0 +1,37 @@
...
+case object GoodsReleasedResponse extends MessageResponse {
+  override val messageReceived: MessageReceived = MessageReceived.GoodsReleased
+  override val messageType: MessageType         = MessageType.GoodsReleased
+  override val xsdFile: XSDFile                 = GoodsReleasedXSD
+}
+case object UnloadingPermissionResponse extends MessageResponse {
+  override val messageReceived: MessageReceived = MessageReceived.UnloadingPermission
+  override val messageType: MessageType         = MessageType.UnloadingPermission
+  override val xsdFile: XSDFile                 = UnloadingPermissionXSD
+}
```
amaalali, line 37: I like that we've taken to model these as objects as opposed to strings floating around the code base 👍 the one downside is that we now seem to have `GoodsReleasedResponse`,  `MessageType.GoodsReleased`, `GoodsReleasedXSD` and `MessageReceived.GoodsReleased`. This here wants to unify some of them, but is grouping it together as a product type with some unrepresentable states, such as a 

 ```
MessageResponse {
  val messageReceived: MessageReceived = MessageReceived.GoodsReleased
  val messageType: MessageType         = MessageType.UnloadingPermission
  val xsdFile: XSDFile                 = ArrivalNotificationRejectionXSD
}
```

I think its a good first solution for this problem,  but we may need to refine it

File: app/controllers/MessageResponseController.scala [hunk 4]
```diff
@@ -0,0 +1,88 @@
...
+                case None =>
+                  logger.error(s"Failure to parse message")
+                  Future.successful(BadRequest)
+              }
+            case Failure(e) =>
+              logger.error(s"Failure to validate against XSD. Exception: ${e.getMessage}")
+              Future.successful(BadRequest)
+          }
+        case None =>
+          Future.successful(BadRequest)
```
amaalali, line 84: We need a named logger here since we're rejecting the incoming message

jhkldev:
Should transition method be removed based on current method of handling MessageResponse.
The transition method appears to be redundant based on how the code is currently working. Either transition can be improved to return an Option or we can remove it.

app/controllers/MessageResponseController.scala
MatthewAlexanderBeck: Does the logger on line 57 - 59 not cover this. We could move it down to this layer if it makes more sense.

conf/app.routes
```diff
@@ -1,12 +1,12 @@
...
+GET                     /movements/arrivals/:arrivalId                          controllers.MovementsController.getArrival(arrivalId: ArrivalId)
+PUT                     /movements/arrivals/:arrivalId                          controllers.MovementsController.putArrival(arrivalId: ArrivalId)
 
-GET                    /movements/arrivals/:arrivalId/messages/:messageId           controllers.MovementsController.getMessage(arrivalId: ArrivalId, messageId: MessageId)
+GET                     /movements/arrivals/:arrivalId/messages/:messageId      controllers.MovementsController.getMessage(arrivalId: ArrivalId, messageId: MessageId)
 
-POST                    /movements/arrivals/:arrivalId/messages                   controllers.MessagesController.post(arrivalId: ArrivalId)
-POST                    /movements/arrivals/:messageSender/goods-released    controllers.GoodsReleasedController.post(messageSender: MessageSender)
+POST                    /movements/arrivals/:messageSender/messages/eis         controllers.MessageResponseController.post(messageSender: MessageSender)
```
chris-ianson, line 11: Like that we're now sticking eis on the end but would prefer this in an eis.routes file that's imported (clear this isn't publicly accessible) Something to keep in mind for future

PR #87, opened 2020-04-27
CTCTRADERS-368: Update headers

MatthewAlexanderBeck:


app/connectors/MessageConnector.scala
```diff
@@ -54,23 +54,19 @@ class MessageConnector @Inject()(config: AppConfig, http: HttpClient)(implicit e
...
-      },
-      "X-Message-Sender" -> messageSender.toString,
-      "X-Forwarded-Host" -> "mdtp",
+      }, // TODO double check this is correct
       "Date"             -> Format.dateFormattedForHeader(dateTime),
-      "Accept"           -> "application/xml"
-      //"Authorization"    -> s"Bearer $bearerToken"
+      "Content-Type"     -> "application/xml",
+      "Accept"           -> "application/xml",
+      "Authorization"    -> "Bearer securityToken", //TODO placeholder, will be provided later
```
chris-ianson, line 68: Double check this works Matt. I think because we have the implicit HMRC headers which already include an Authorization header, when you add this in it duplicates it. It works locally but when you deploy into env Nginx throws a 400 error (the original problem we found - after a week of searching - a couple of months back)

amaalali: Good catch, I think play's `Headers` class has a way to remove headers and then we can add this in. Does that sound like a better solution?

MatthewAlexanderBeck: Thanks for the spot guys. I think we can override this (rather than removing just to add in again) with following below: 

` val newHeaders = headerCarrier
      .copy(authorization = Some(Authorization("Bearer securityToken"))) //TODO placeholder, will be provided later
      .withExtraHeaders(addHeaders(message.messageType, dateTime, messageSender): _*)`

Do you think this is a sensible solution to the Nginx error?

chris-ianson: That looks like it would work

app/connectors/MessageConnector.scala
```diff
@@ -47,7 +48,9 @@ class MessageConnector @Inject()(config: AppConfig, http: HttpClient)(implicit e
 
     lazy val messageSender = MessageSender(arrivalId, message.messageCorrelationId)
 
-    val newHeaders = headerCarrier.withExtraHeaders(addHeaders(message.messageType, dateTime, messageSender): _*)
+    val newHeaders = headerCarrier
+      .copy(authorization = Some(Authorization("Bearer securityToken"))) //TODO placeholder, will be provided later
```
amaalali, line 52: We can remove the TODO for now since it's tracked by a ticket
```suggestion
      .copy(authorization = Some(Authorization("Bearer securityToken")))
```

PR #91, opened 2020-04-29
Fix compilation warning

ashwinirudrappa:


app/repositories/ArrivalMovementRepository.scala
```diff
@@ -202,15 +203,26 @@ class ArrivalMovementRepository @Inject()(mongo: ReactiveMongoApi)(implicit ec:
...
-              le =>
-                if (le.updatedExisting) Success(()) else Failure(new Exception(s"Could not find arrival $arrivalId"))
-            }
-            .getOrElse(Failure(new Exception("Failed to update arrival")))
-        }
+      _.findAndUpdate(
+        selector = selector,
+        update = modifier,
+        fetchNewObject = true,
+        upsert = true,
```
amaalali, line 210: This isn't valid, it should fail if this happens

app/repositories/ArrivalMovementRepository.scala
```diff
@@ -231,8 +243,19 @@ class ArrivalMovementRepository @Inject()(mongo: ReactiveMongoApi)(implicit ec:
       )
 
     collection.flatMap {
-      _.findAndUpdate(selector, modifier)
-        .map {
+      _.findAndUpdate[JsObject, JsObject](
+        selector = selector,
+        update = modifier,
+        fetchNewObject = true,
+        upsert = true,
```
amaalali, line 250: Same here, upsert should be false

ashwinirudrappa: it was true earlier,just used what ever was set

PR #94, opened 2020-04-30
Change getMessageType check in MessagesController

jhkldev:


app/controllers/MessagesController.scala
```diff
@@ -67,6 +65,8 @@ class MessagesController @Inject()(
                   }
             }
             .getOrElse(Future.successful(BadRequest("Invalid data: missing either DatOfPreMES9, TimOfPreMES10 or DocNumHEA5")))
+        case _ =>
+          Future.successful(NotImplemented)
```
amaalali, line 69: Can we add a quick test for this branch? It will prompt anyone to remove this when they add support for other messages

jhkldev: There is a test for this in the existing code https://github.com/hmrc/transit-movements-trader-at-destination/blob/master/test/controllers/MessagesControllerSpec.scala#L215

PR #96, opened 2020-04-30
Ctctraders 446

kevincoatsworth:


test/models/ArrivalStatusSpec.scala
```diff
@@ -34,6 +34,10 @@ class ArrivalStatusSpec extends SpecBase with ScalaCheckDrivenPropertyChecks wit
       Initialized.transition(MessageReceived.GoodsReleased) mustEqual GoodsReleased
     }
 
+    "to Goods Rejection when receiving a Goods Rejection event" in {
+      Initialized.transition(MessageReceived.GoodsRejected) mustEqual GoodsRejected
+    }
+
```
amaalali, line 40: This may not be a valid transition, we need to have a discussion about if this should be validated against

PR #99, opened 2020-05-04
Added GetMessages for Arrival Endpoint

RobertLaverick:


app/controllers/MessagesController.scala
```diff
@@ -69,4 +76,26 @@ class MessagesController @Inject()(
...
+    implicit request =>
+      if (request.arrival.messages.isDefinedAt(messageId.index) && request.arrival.messages(messageId.index).optStatus != Some(SubmissionFailed))
+        Ok(Json.toJson(ResponseMovementMessage.build(arrivalId, messageId, request.arrival.messages(messageId.index))))
+      else NotFound
+  }
+
+  def getMessages(arrivalId: ArrivalId): Action[AnyContent] = authenticateForRead(arrivalId) {
+    implicit request =>
+      Ok(
+        Json.toJson(ResponseArrival.build(
```
amaalali, line 90: Can we use `Json.toJsObject`? This may force us to define ResponseArrival and ResponseMessages as OWrites

app/controllers/MessagesController.scala
```diff
@@ -69,4 +76,26 @@ class MessagesController @Inject()(
...
+          request.arrival,
+          request.arrival.messages.view.zipWithIndex
+            .filterNot {
+              case (message, _) => message.optStatus == Some(SubmissionFailed)
+            }
+            .map {
+              case (message, count) => ResponseMovementMessage.build(arrivalId, new MessageId(count + 1), message)
+            }
```
amaalali, line 98: I think this is logic that can be moved out of the controller and should be done by something else, possibly by whatever is doing the conversion from the Arrival to the ResponseArrival?

app/controllers/MessagesController.scala
```diff
@@ -69,4 +76,26 @@ class MessagesController @Inject()(
...
+
+  def getMessages(arrivalId: ArrivalId): Action[AnyContent] = authenticateForRead(arrivalId) {
+    implicit request =>
+      Ok(
+        Json.toJson(ResponseArrival.build(
+          request.arrival,
+          request.arrival.messages.view.zipWithIndex
+            .filterNot {
+              case (message, _) => message.optStatus == Some(SubmissionFailed)
+            }
```
jhkldev, line 95: Should we return only submission pending and submission succeeded messages?

PR #100, opened 2020-05-04
Change lock created and message dateTime to save as Dates in Mongo

amaalali:


it/repositories/ArrivalMovementRepositorySpec.scala
```diff
@@ -34,7 +34,18 @@ class ArrivalMovementRepositorySpec
...
+  override def afterAll(): Unit = {
+    super.afterAll()
+    database.flatMap {
+      db =>
+        val arrivalMovement1 = arbitrary[Arrival].sample.value
+        val arrivalMovement2 = arbitrary[Arrival].sample.value
+
+        db.collection[JSONCollection](ArrivalMovementRepository.collectionName).insert(false).many(Seq(arrivalMovement1, arrivalMovement2))
+    }.futureValue
+  }
```
MatthewAlexanderBeck, line 48: Why do we need to insert two new documents after the test suite is run? Isnt afterAll mostly used for cleanup purposes?

amaalali: Will remove, was used while debugging

PR #102, opened 2020-05-05
Split the message status events and the arrival movement events

amaalali:
These two statuses have been decoupled since initial design but there is still a dependency of the Arrival Status on the Message status, but not the other way.

This change splits the two to be more indicative of the independence of the types.


app/services/SaveMessageService.scala
```diff
@@ -20,7 +20,9 @@ import com.google.inject.Inject
 import models.ArrivalStatus
 import models.MessageResponse
 import models.MessageSender
-import models.SubmissionResult
+import models.SubmissionProcessingResult
+import models.SubmissionProcessingResult
+import models.SubmissionProcessingResult._
```
MatthewAlexanderBeck, line 25: Super minor but this is repeated

PR #105, opened 2020-05-07
CTCTRADERS-400 

amaalali:
Limit the adding of 007 to existing movement to only when there are failed 007 messages

### DIFF
--- a/app/controllers/MovementsController.scala
+++ b/app/controllers/MovementsController.scala
@@ -21,14 +21,16 @@ import controllers.actions.AuthenticatedGetArrivalForReadActionProvider
 import controllers.actions.AuthenticatedGetArrivalForWriteActionProvider
 import controllers.actions.AuthenticatedGetOptionalArrivalForWriteActionProvider
 import javax.inject.Inject
+import models.MessageStatus.SubmissionFailed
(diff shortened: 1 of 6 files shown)


app/controllers/MovementsController.scala
```diff
@@ -56,10 +58,13 @@ class MovementsController @Inject()(
 )(implicit ec: ExecutionContext)
     extends BackendController(cc) {
 
+  private val atLeastOneMessageSent: Seq[MovementMessage] => Boolean =
+    _.exists(x => x.optStatus.contains(SubmissionSucceeded) || x.optStatus.isEmpty)
```
jhkldev, line 62: Can the x.optStatus.isEmpty condition be removed?

amaalali: No, if there is a response from upstream we can shortcircuit out of the list traversal

RobertLaverick: Shouldn't the x.optStatus.isEmpty be first in this condition? I'm not sure how it'll handle contains if it is empty.

amaalali: `None.contains` will return false

PR #109, opened 2020-05-13
CTDA 211: Add debugging when returning BadRequest

jhkldev:


app/controllers/NCTSMessageController.scala
```diff
@@ -57,14 +57,23 @@ class NCTSMessageController @Inject()(cc: ControllerComponents, getArrival: GetA
...
-          saveMessageService.validateXmlAndSaveMessage(xml, messageSender, response, newState) map {
-            case SubmissionSuccess         => Ok
-            case SubmissionFailureInternal => InternalServerError
-            case SubmissionFailureExternal => BadRequest
+          val newState         = request.arrival.status.transition(response.messageReceived)
+          val processingResult = saveMessageService.validateXmlAndSaveMessage(xml, messageSender, response, newState)
+          processingResult map {
+            case SubmissionSuccess => Ok
+            case SubmissionFailureInternal =>
+              val message = "Internal Submission Failure " + processingResult
```
amaalali, line 65: The `processingResult ` isn't going to give much information here aside from if it was a `SubmissionFailureInternal` vs a `SubmissionFailureExternal`, but if that's what's wanted then its fine

PR #110, opened 2020-05-13
Change the Arrival messages to be nonempty

amaalali:
This changes the Arrival model to not allow the messages to be empty and removes logic that handles the empty messages which is not possible.


app/controllers/MovementsController.scala
```diff
@@ -55,7 +56,7 @@ class MovementsController @Inject()(
 )(implicit ec: ExecutionContext)
     extends BackendController(cc) {
 
-  private val allMessageUnsent: Seq[MovementMessage] => Boolean =
+  private val allMessageUnsent: NonEmptyList[MovementMessage] => Boolean =
```
chris-ianson, line 59: What's this used for?

app/models/Arrival.scala
```diff
@@ -27,12 +30,30 @@ case class Arrival(
...
   updated: LocalDateTime,
-  messages: Seq[MovementMessage],
+  messages: NonEmptyList[MovementMessage],
   nextMessageCorrelationId: Int
 )
 
 object Arrival {
 
+  @deprecated("use NonEmptyList[MovementMessage] over Seq[MovementMessage]", "now")
+  def apply2(arrivalId: ArrivalId,
```
chris-ianson, line 40: How come we're marking this as deprecated and not just changing it to use a NonEmptyList?

amaalali: Oops this was while I was trying to get back to a compiling state, I'll get rid of that

app/controllers/MovementsController.scala
amaalali: This is the check that determines if a movement is will end up being created as a brand new or as part of a previous movement in which all messages (IE007) are unsubmitted

app/models/Arrival.scala
amaalali: Removed

PR #112, opened 2020-05-14
Add custom error handler to log client and server errors

jhkldev:


jhkldev:
> I think you ought to make this extend the uk.gov.hmrc.play.bootstrap.JsonErrorHandler rather than simply replacing it.
> 
> However I'd 100% ensure you get @amaalali to review this

I think extending JsonErrorHandler will work only with requests which have an Accept header with json as the value. I want to log all types of requests.

PR #114, opened 2020-05-20
added get arrival endpoint

RobertLaverick:


app/controllers/actions/GetArrivalAction.scala
```diff
@@ -68,13 +69,19 @@ private[actions] class AuthenticatedGetArrivalAction(
...
+          Right(ArrivalRequest(request.request, arrival))
+        case Some(_) =>
+          Logger.warn("Attempt to retrieve an arrival for another EORI")
+          Left(NotFound)
+        case None =>
+          Left(NotFound)
+      }
+      .recover {
+        case e =>
+          Left(InternalServerError(s"Failed with the following error: $e"))
```
amaalali, line 85: I don't think we should return this error as part of the response but it makes sense to log it

app/controllers/MovementsController.scala
```diff
@@ -144,7 +146,10 @@ class MovementsController @Inject()(
         }
   }
 
-  def getArrival(arrivalId: ArrivalId): Action[AnyContent] = defaultActionBuilder(_ => NotImplemented)
+  def getArrival(arrivalId: ArrivalId): Action[AnyContent] = authenticatedArrivalForRead(arrivalId) {
+    implicit request =>
+      Ok(Json.toJsObject(request.arrival))
```
amaalali, line 151: This won't be the correct json response which would be using this writes: https://github.com/hmrc/transit-movements-trader-at-destination/blob/master/app/models/Arrival.scala#L58-L68

app/controllers/actions/GetArrivalAction.scala
RobertLaverick: I'll log it as an error

app/controllers/MovementsController.scala
RobertLaverick: That's fair - in which case this will also apply for the getArrivals endpoint. Do you want me to apply the ResponseArrival model to that as well?

app/models/response/ResponseArrivalWithMessages.scala
```diff
@@ -0,0 +1,62 @@
...
+      arrival.created,
+      arrival.updated,
+      arrival.messages.toList.zipWithIndex
+        .filterNot {
+          case (message, _) => message.optStatus == Some(SubmissionFailed)
+        }
+        .map {
+          case (message, count) => ResponseMovementMessage.build(arrival.arrivalId, new MessageId(count + 1), message)
+        }
+    )
```
amaalali, line 58: I believe there is an update in the master branch that simplifies this

app/controllers/MovementsController.scala
```diff
@@ -143,15 +146,21 @@ class MovementsController @Inject()(
...
     implicit request =>
       arrivalMovementRepository
         .fetchAllArrivals(request.eoriNumber)
         .map {
           allArrivals =>
-            Ok(Json.toJsObject(Arrivals(allArrivals)))
+            Ok(Json.toJsObject(ResponseArrivals(allArrivals.map {
+              arrival =>
+                ResponseArrival.build(arrival)
+            })))
```
amaalali, line 163: Good spot on this one 👍

PR #115, opened 2020-05-20
CTCTRADERS-447 Add service that provides summary of the messages in an Arrival

amaalali:
This only adds support for the IE007 and IE008 and further work will need to add support for the IE043, and IE044, but this adds the first design of how we work these out

### DIFF
--- a/app/services/ArrivalMessageSummaryService.scala
+++ b/app/services/ArrivalMessageSummaryService.scala
@@ -42,24 +42,32 @@ class ArrivalMessageSummaryService {
         .toList
         .maxBy(_._1.messageCorrelationId)
 
-    case _ => ??? // Unreachable but unprovable
+    case NonEmptyList((msg, _), _) =>
+      // Unreachable but unprovable
+      throw new RuntimeException(
+        "Reached an invalid state when summarizing Arrival Notification. " +
+          "Expected the first message of the movement to be MovementMessageWithStatus with an ArrivalNotification, " +
+          s"but got ${msg.getClass} that contained a ${msg.messageType.code}"
+      )
   }
 
   def arrivalRejection(arrival: Arrival): Option[(MovementMessage, MessageId)] = {
 
-    lazy val numIe007 = arrival.messages.toList.count {
+    lazy val arrivalNotificationCount = arrival.messages.toList.count {
       case MovementMessageWithStatus(_, ArrivalNotification, _, _, _) => true
       case _                                                          => false
     }
 
-    val ie008Messages = arrival.messagesWithId
+    val rejectionNotifications = arrival.messagesWithId
       .foldLeft(Seq.empty[(MovementMessageWithoutStatus, MessageId)]) {
         case (acc, (m @ MovementMessageWithoutStatus(_, ArrivalRejection, _, _), mid)) => acc :+ Tuple2(m, mid)
         case (acc, _)                                                                  => acc
       }
 
-    if (ie008Messages.nonEmpty && numIe007 == ie008Messages.length)
-      Some(ie008Messages.maxBy(_._1.messageCorrelationId))
+    val rejectionNotificationCount = rejectionNotifications.length
+
+    if (rejectionNotificationCount > 0 && arrivalNotificationCount == rejectionNotificationCount)
+      Some(rejectionNotifications.maxBy(_._1.messageCorrelationId))
     else
       None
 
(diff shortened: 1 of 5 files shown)


app/services/ArrivalMessageSummaryService.scala
```diff
@@ -0,0 +1,68 @@
...
+
+      tail
+        .foldLeft(NonEmptyList.of(head))({
+          case (acc, (m @ MovementMessageWithStatus(_, ArrivalNotification, _, _, _), mid)) => acc :+ Tuple2(m, mid)
+          case (acc, _)                                                                     => acc
+        })
+        .toList
+        .maxBy(_._1.messageCorrelationId)
+
+    case _ => ??? // Unreachable but unprovable
```
garethgooch, line 45: even if its unreachable shouldn't it do something?

test/services/SaveMessageServiceSpec.scala
```diff
@@ -141,7 +141,7 @@ class SaveMessageServiceSpec extends SpecBase with BeforeAndAfterEach {
     }
 
     "return Failure when we cannot parse the message due malformed time" in {
-      when(mockXmlValidationService.validate(any(), any())).thenReturn(Success())
+      when(mockXmlValidationService.validate(any(), any())).thenReturn(Success(()))
```
garethgooch, line 144: is there a reason for adding in the unit here? xmlValidationService wasn't touched

app/services/ArrivalMessageSummaryService.scala
```diff
@@ -0,0 +1,68 @@
...
+      case _                                                          => false
+    }
+
+    val ie008Messages = arrival.messagesWithId
+      .foldLeft(Seq.empty[(MovementMessageWithoutStatus, MessageId)]) {
+        case (acc, (m @ MovementMessageWithoutStatus(_, ArrivalRejection, _, _), mid)) => acc :+ Tuple2(m, mid)
+        case (acc, _)                                                                  => acc
+      }
+
+    if (ie008Messages.nonEmpty && numIe007 == ie008Messages.length)
```
chris-ianson, line 61: Maybe some renaming of values would improve readability here. ie008Messages.nonEmpty could be _hasArrivalRejectionMessage_ (with the nonEmpty in the implementation)

numIe007 could be _arrivalNotificationCount_
ie008Messages could be _rejectionNotificationCount_

test/services/ArrivalMessageSummaryServiceSpec.scala
```diff
@@ -0,0 +1,219 @@
...
+              messageId mustEqual MessageId.fromMessageIdValue(1).value
+          }
+      }
+
+    }
+
+    "must return the new IEOO7 when there has been a correction to a rejected arrival message" in {
+      val service = new ArrivalMessageSummaryService
+
+      forAll(ie007Gen.submitted.msgCorrId(1), ie008Gen.msgCorrId(1), ie007Gen.msgCorrId(2)) {
```
chris-ianson, line 97: Does this mean we can set values manually here rather than taking the generated value and overriding values with _.copy()_

test/services/SaveMessageServiceSpec.scala
amaalali: To resolve compiler warnings

test/services/ArrivalMessageSummaryServiceSpec.scala
amaalali: Under the hood it does still call copy, it just makes it easier to write, https://github.com/hmrc/transit-movements-trader-at-destination/pull/115/files/f8728c3b88ace423fe860c14d275e9330acc07f4#diff-ae0ea5187d42d5f23a9b45db82935870R209-R211

app/services/ArrivalMessageSummaryService.scala
amaalali: Throwing an exception would be the only other thing, but the question is what sort of exception?

Possibly a RuntimeException, but that doesn't seem informative enough, it more that we hit a scenario that we shouldn't ever be able to get to 😕 

Any suggestions?

amaalali: The only thing we can do here would possibly be to throw an exception (until someone makes this better).

I'm not sure a RuntimeException is correct here 🤔 we want the exception to indicate that we've hit a scenario that we shouldn't be able to. Any suggestions?

test/services/SaveMessageServiceSpec.scala
amaalali: To resolve compiler warnings

test/services/ArrivalMessageSummaryServiceSpec.scala
chris-ianson: Nice one 👍

PR #117, opened 2020-05-21
Add endpoint to get summary of arrival messages

amaalali:


app/controllers/MessagesSummaryController.scala
```diff
@@ -0,0 +1,49 @@
...
+import play.api.mvc.Action
+import play.api.mvc.AnyContent
+import play.api.mvc.ControllerComponents
+import services.ArrivalMessageSummaryService
+import uk.gov.hmrc.play.bootstrap.controller.BackendController
+
+import scala.concurrent.ExecutionContext
+
+class MessagesSummaryController @Inject()(
+  authenticateForRead: AuthenticatedGetArrivalForReadActionProvider,
```
MatthewAlexanderBeck, line 33: Do you think it is worth potentially splitting this action up into two? maybe have an authenticate action and a getArrival action?

MatthewAlexanderBeck: ^^ this would make it consistent with what we do on the frontend where we chain actions on the controller

amaalali: We went for the design of composing these actions around the use cases for which we have 3 for now:
- `AuthenticatedGetArrivalForReadActionProvider`
- `AuthenticatedGetArrivalForWriteActionProvider`
- `GetArrivalForWriteActionProvider`

There is one more, but that isn't generic and is specialized to the one controller where it's used, it's not reusable really so ignore it for this convo. It may actually technical debt but not one that we know how to act on as yet
- `AuthenticatedGetOptionalArrivalForWriteActionProvider`

amaalali: We went for the design of composing these actions around the use cases for which we have 3 for now:
- we're not modifying the arrival movement: `AuthenticatedGetArrivalForReadActionProvider`
- we need to  modify the arrival movement based on user action: `AuthenticatedGetArrivalForWriteActionProvider`
- we need to  modify the arrival movement based on responses from core: `GetArrivalForWriteActionProvider`

There is one more, but that isn't generic and is specialized to the one controller where it's used, it's not reusable really so ignore it for this convo. It may actually technical debt but not one that we know how to act on as yet
- `AuthenticatedGetOptionalArrivalForWriteActionProvider`

But based on the use cases

MatthewAlexanderBeck: I think this just comes down to personal preference as I prefer the decomposed actions in the controller, it makes it easier to understand what actions are required without having to jump into another action provider first however it isnt really too much difference so I will go ahead and approve :+1:

PR #118, opened 2020-05-22
MessageSummaryService now exposes single interface that returns a summary

amaalali:


app/controllers/actions/AuthenticateActionProvider.scala
```diff
@@ -22,11 +22,15 @@ import play.api.mvc.ActionBuilder
...
+/*
+  This is made available for testing purposes. Do not extend this,
+  use the other actions that provide the behaviour that is needed.
+ */
 trait AuthenticateActionProvider {
   def apply(): ActionBuilder[AuthenticatedRequest, AnyContent]
 }
 
-class AuthenticateActionProviderImpl @Inject()(
+private[actions] class AuthenticateActionProviderImpl @Inject()(
```
MatthewAlexanderBeck: This breaks the Module class

app/controllers/actions/AuthenticateActionProvider.scala
```diff
@@ -22,11 +22,15 @@ import play.api.mvc.ActionBuilder
...
   authenticate: AuthenticateAction,
   buildDefault: DefaultActionBuilder
 ) extends AuthenticateActionProvider {
```
MatthewAlexanderBeck: Could this be moved into Spec or a testOnly folder if only made for testing?

amaalali: Ah sorry, the trait is only made available so that we can create a fake, but the `AuthenticateActionProviderImpl` is what should be used

app/controllers/actions/AuthenticateActionProvider.scala
amaalali: Reverted for this PR

PR #120, opened 2020-06-05
CTC-TRADERS-477 Update MRN on PUT /movements/arrivals/:arrivalId

amaalali:
This PR addresses the primary concern that on `PUT /movements/arrivals/:arrivaId`, cached data values (for now this is MRN) were not updated.

The second concern was that the repository update methods, for example, setState, setMessageState and setArrivalStateAndMessageState were all highly specialised to update certain fields in the document and were not composable or reusable outside of those use cases. Furthermore, they all had different return values and their implementations were subtly different despite doing the same type of operation. This second concern is addressed by adding a new public method that is meant to replace all of those methods by providing a way to generically describe and update on an arrival.


test/models/ArrivalUpdateSpec.scala
```diff
@@ -0,0 +1,179 @@
...
+      }
+
+      "is not defined when both message status are not defined" in {
+        forAll(noMessageUpdate(), noMessageUpdate()) {
+          (lhs, rhs) =>
+            lhs.combine(rhs).messageUpdate must not be (defined)
+        }
+      }
+
+      "THIS MAKES IT A MONIOD" ignore {
```
ashwinirudrappa, line 107: @amaalali Is this test supposed to be ignored

test/controllers/actions/GetArrivalForWriteActionProviderSpec.scala
```diff
@@ -41,8 +41,8 @@ import repositories.LockRepository
 import scala.concurrent.Future
 
 class GetArrivalForWriteActionProviderSpec
-    extends FreeSpec
-    with MustMatchers
+    extends AnyFreeSpec
```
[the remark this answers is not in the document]
amaalali: It seems that in the newer version of ScalaTest they've depreciated `FreeSpec`, renamed it to `AnyFreeSpec` and moved it to a different sub-package

test/models/ArrivalUpdateSpec.scala
amaalali: Whoops, wasn't supposed to commit that, I'll remove it 😬

app/controllers/MovementsController.scala
```diff
@@ -124,21 +124,22 @@ class MovementsController @Inject()(
...
         .map {
-          submitMessageService
-            .submitMessage(arrivalId, request.arrival.nextMessageId, _, ArrivalStatus.ArrivalSubmitted)
-            .map {
-              case SubmissionProcessingResult.SubmissionSuccess =>
-                Accepted("Message accepted")
-                  .withHeaders("Location" -> routes.MovementsController.getArrival(request.arrival.arrivalId).url)
+          case (message, mrn) =>
+            submitMessageService
+              .submitIe007Message(arrivalId, request.arrival.nextMessageId, message, mrn)
```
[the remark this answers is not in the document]
amaalali: I don't particularly like this style, I think that `SubmitMessageService#submitMessage` and `SubmitMessageService#submitIe007Message` need to be combined into one method. I don't think that `submitMessage` should take in the ArrivalState but should work it out using the transition on the `ArrivalStatus` and the `MessageReceivedEvent`

app/controllers/MovementsController.scala
```diff
@@ -124,21 +124,22 @@ class MovementsController @Inject()(
   def putArrival(arrivalId: ArrivalId): Action[NodeSeq] = authenticateForWrite(arrivalId).async(parse.xml) {
```
chris-ianson, line 124: Do we need _putArrival_? We're in the movements controller so _put_ might be enough to indicate we're creating a new movement?

app/models/ArrivalUpdate.scala
```diff
@@ -0,0 +1,63 @@
...
+            )
+      )
+    )
+}
+
+case class ArrivalUpdate(arrivalUpdate: Option[ArrivalStatus], messageUpdate: Option[MessageStatusUpdate])
+
+object ArrivalUpdate {
+
+  implicit object ArrivalUpdateSemiGroupInst extends Semigroup[ArrivalUpdate] {
```
chris-ianson, line 45: Semigroup?

app/services/SubmitMessageService.scala
```diff
@@ -79,6 +83,50 @@ class SubmitMessageService @Inject()(
       }
     }
 
+  def submitIe007Message(arrivalId: ArrivalId, messageId: MessageId, message: MovementMessageWithStatus, mrn: MovementReferenceNumber)(
```
chris-ianson, line 86: Are submitMessage and submitIe007Message doing the same thing?

test/base/FreeSpecDiscipline.scala
```diff
@@ -0,0 +1,37 @@
...
+
+import org.scalactic.Prettifier
+import org.scalactic.source.Position
+import org.scalatest.freespec.AnyFreeSpecLike
+import org.scalatest.prop.Configuration
+import org.scalatestplus.scalacheck.Checkers
+import org.typelevel.discipline.Laws
+import org.typelevel.discipline.scalatest.Discipline
+
+trait FreeSpecDiscipline extends Discipline {
```
chris-ianson, line 27: How does this work?

test/controllers/actions/GetArrivalForWriteActionProviderSpec.scala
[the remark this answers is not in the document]
chris-ianson: So if we update ScalaTest we'll need to update the FreeSpec in other repos?

app/controllers/MovementsController.scala
amaalali: I think that at the time we didn't spilt the controllers by resource so it made sense then, but you're right it might be unnecessary now, I'll probably raise a PR to rename that

app/services/SubmitMessageService.scala
amaalali: They are 😞
This was duplicated because this method is more generic, but I don't think that its the final design. The main limitation that made this a necessary, but temporary 🤞, duplication is the repository methods that we call. Those methods are highly specialized to the use cases and don't have a way to abstract over them. 

The Semigroup is an attempt to do that so that you have different types of updates that you give to the repository layer and you can combine them together and ask the repository layer to do its thing using the combined updates (or single update type).

test/base/FreeSpecDiscipline.scala
amaalali: The Discipline trait has a `checkAll` method that you implement for your testing style and it runs the `Laws#RuleSet` that you pass in. There are a few implementations for other testing styles but there wasn't any FreeSpec. The RuleSet hooks in with ScalaCheck to run the tests and uses ScalaCheck's `forAll.

app/models/ArrivalUpdate.scala
amaalali: I've started modelling all the updates that can happen as a `Semigroup`, which means that we can combine them together. So given two `ArrivalUpdate`, we can combine them together to get an `ArrivalUpdate` that (in our case) is a combination of the two. Using cats here means that I get a few extra things that we don't need to do manually like it being associative and we can use the cats-laws for some testing.

An example of how we can use this is given an ArrivalUpdate to the Arrival's state, and an ArrivalUpdate to a Message's state, we can combine them together to get an update type that captures both these operations, and then we can pass that along to be interpreted by repository layer. We don't leverage this yet, we still need to move those out the repository layer and get an intermediate layer to construct those ArrivalUpdates for us

PR #124, opened 2020-06-17
Refactor representation of ArrivalUpdate

amaalali:
- Model the arrival selectors as objects
- Remove all uses of `setMessageState` and use `updateArrival`
- Refactor ArrivalUpdate using Scala 2.12 fix to implicit lookup which allows implicit resolution to use `F[A]` for an `AA` subtype of `A`

Outstanding:

- SubmitMessageService is still doing too much, the logic to determine the updates should probably be moved to something else. This would allow us to increase the testing of this and make the system better tested.


app/models/ArrivalUpdate.scala
```diff
@@ -17,11 +17,44 @@
...
+sealed trait ArrivalUpdate
+
+object ArrivalUpdate {
+
+  implicit val arrivalUpdateSemigroup: Semigroup[ArrivalUpdate] = {
+    case (ArrivalStatusUpdate(_), x @ ArrivalStatusUpdate(_))        => x
+    case (a @ ArrivalStatusUpdate(_), m @ MessageStatusUpdate(_, _)) => CompoundStatusUpdate(a, m)
+    case (ArrivalStatusUpdate(_), c @ CompoundStatusUpdate(_, _))    => c
+    case (ArrivalStatusUpdate(_), p @ ArrivalPutUpdate(_, _))        => p
+
```
MatthewAlexanderBeck, line 32: Are there any plans on using the params? As I feel the Object(_, _) format makes this difficult to read compared to something like:

(Case _: ArrivalStatusUpdate, _: ArrivalStatusUpdate)

app/models/ArrivalUpdate.scala
```diff
@@ -17,11 +17,44 @@
...
+    case (p @ ArrivalPutUpdate(_, _), a @ ArrivalStatusUpdate(_))     => p.copy(arrivalUpdate = p.arrivalUpdate.copy(arrivalStatusUpdate = a))
+    case (p @ ArrivalPutUpdate(_, _), m @ MessageStatusUpdate(_, _))  => p.copy(arrivalUpdate = p.arrivalUpdate.copy(messageStatusUpdate = m))
+    case (p @ ArrivalPutUpdate(_, _), c @ CompoundStatusUpdate(_, _)) => p.copy(arrivalUpdate = c)
+  }
+
+  implicit val arrivalUpdateArrivalModifier: ArrivalModifier[ArrivalUpdate] = {
+    case x @ MessageStatusUpdate(_, _)  => ArrivalModifier.toJson(x)
+    case x @ ArrivalStatusUpdate(_)     => ArrivalModifier.toJson(x)
+    case x @ CompoundStatusUpdate(_, _) => ArrivalModifier.toJson(x)
+    case x @ ArrivalPutUpdate(_, _)     => ArrivalModifier.toJson(x)
```
MatthewAlexanderBeck, line 53: Is this matcher needed if the outcome is always the same?

test/models/ArrivalUpdateSpec.scala
```diff
@@ -19,153 +19,131 @@ package models
...
-        forAll(messageStatusUpdate, messageStatusUpdate) {
-          (lhs, rhs) =>
-            lhs.combine(rhs).messageUpdate mustEqual rhs.messageUpdate
-        }
-      }
+          val expectedValue =
+            if (rhs.isInstanceOf[MessageStatusUpdate] || rhs.isInstanceOf[CompoundStatusUpdate] || rhs.isInstanceOf[ArrivalPutUpdate])
+              removeMessageUpdate(ArrivalModifier.toJson(lhs)) deepMerge ArrivalModifier.toJson(rhs)
+            else
+              ArrivalModifier.toJson(lhs) deepMerge ArrivalModifier.toJson(rhs)
```
MatthewAlexanderBeck, line 68: With this bit in mind do we potentially need to split this into two tests?

app/models/ArrivalUpdate.scala
amaalali: This is allow implicit resolution to pick the correct value for the ArrivalModifier.toJson

app/models/ArrivalUpdate.scala
```diff
@@ -39,26 +72,35 @@ object MessageStatusUpdate extends MongoDateTimeFormats {
...
-      arrivalUpdateJson deepMerge messageUpdateJson
-    }
-  }
+object ArrivalPutUpdate extends MongoDateTimeFormats {
+
+  implicit val arrivalPutUpdateArrivalModifier: ArrivalModifier[ArrivalPutUpdate] = a =>
+    Json.obj(
+      "$set" -> Json.obj(
+        "movementReferenceNumber" -> a.movementReferenceNumber,
+        "lastUpdated"             -> LocalDateTime.now.withSecond(0).withNano(0)
```
MatthewAlexanderBeck, line 102: Getting the LocalDateTime.now but manually setting the seconds and nano to 0. Are these required as we can set a formatter to remove them?

it/repositories/ArrivalMovementRepositorySpec.scala
```diff
@@ -108,21 +109,23 @@ class ArrivalMovementRepositorySpec
 
         val app: Application = builder.build()
 
-        val arrivalStatus: ArrivalStatus = Initialized
-        val arrival                      = arrivalWithOneMessage.sample.value.copy(status = GoodsReleased)
-        val latsUpdated                  = LocalDateTime.now.withSecond(0).withNano(0)
+        val arrivalStatus = ArrivalStatusUpdate(Initialized)
+        val arrival       = arrivalWithOneMessage.sample.value.copy(status = GoodsReleased)
+        val latsUpdated   = LocalDateTime.now.withSecond(0).withNano(0)
```
MatthewAlexanderBeck, line 114: Same as above ^^

app/models/ArrivalUpdate.scala
amaalali: I think that the setting the seconds and nano to 0 is a workaround for testing purposes, which is a bad practice. I'll raise a tech debt issue for this

it/repositories/ArrivalMovementRepositorySpec.scala
amaalali: See above

PR #126, opened 2020-06-29
Ctctraders 441 r:Remove movements older than 28 days from backend

ashwinirudrappa:


app/repositories/ArrivalMovementRepository.scala
```diff
@@ -106,9 +124,12 @@ class ArrivalMovementRepository @Inject()(mongo: ReactiveMongoApi)(implicit ec:
 
     import models.ArrivalModifier.toJson
 
+    val setLastUpdated         = Json.obj("$set" -> Json.obj("lastUpdated" -> LocalDateTime.now))
```
amaalali, line 127: I think we should move this outside in the ArrivalModifier instance

app/repositories/ArrivalMovementRepository.scala
```diff
@@ -64,12 +79,15 @@ class ArrivalMovementRepository @Inject()(mongo: ReactiveMongoApi)(implicit ec:
   private def collection: Future[JSONCollection] =
     mongo.database.map(_.collection[JSONCollection](collectionName))
 
-  def insert(arrival: Arrival): Future[Unit] =
+  def insert(arrival: Arrival): Future[Unit] = {
+    val updatedArrival = arrival copy (lastUpdated = LocalDateTime.now)
```
amaalali, line 83: Since you've defaulted the constructor on the `Arrival` model to be `LocalDateTime.now` is this necessary?

app/repositories/ArrivalMovementRepository.scala
ashwinirudrappa: @amaalali Ya started doing in ArrivalModifier, there were few concerns around like unit testing and have to place the lastUpdated code in All the different modifiers. Placing it in here helps in future also if we bring some new modifiers we no need to worry about 'lastUpdated'

app/repositories/ArrivalMovementRepository.scala
ashwinirudrappa: @amaalali This is just extra protection and to make us more aware when we look into it

amaalali: In that case we can remove it, it's the same as the Arrival status field, which should be `Initialised` at this point but that isn't the responsibility of this layer to enforce

app/repositories/ArrivalMovementRepository.scala
amaalali: I understand what you're getting at, but this seems to break what ArrivalUpdate and ArrivalModifier is trying to encapsulate, you now need to look in multiple places to see what changes are being applied to a movement

PR #127, opened 2020-06-29
CTCTRADERS-573: Save IE058 message to movement in backend

kenneth-rayner:


kenneth-rayner:
messaged @fred-jackson for content and was referred to following attachment linked to ticket 431:
https://docs.google.com/document/d/1h4f1WKWhZmR-asFpNLPsrZHytQIzimN1rgLqyTcZQ44/edit

chris-ianson:
@amaalali @MatthewAlexanderBeck Copying you both into this PR. l know you've both worked on the NCTSController (I think?!) Refactored it to move some of the header validation logic into an action transformer and thin down the controller. If you could have a look and let me know what you think I'd appreciate it.

app/controllers/actions/InboundMessageTransformer.scala
```diff
@@ -0,0 +1,71 @@
...
+        val nextState = request.arrival.status.transition(response.messageReceived)
+
+        Future.successful(
+          new InboundRequest(Some(MessageInbound(response, nextState)), request)
+        )
+      case None =>
+        Future.successful(
+          new InboundRequest(None, request)
+        )
+    }
```
amaalali, line 62: I'm wondering if making this act more as an ActionFilter would be better? So for unsupported message types we send back a bad request here? As it stands this None is being dealt with by the controller directly

app/controllers/actions/InboundMessageTransformer.scala
```diff
@@ -0,0 +1,71 @@
...
+class InboundMessageTransformer @Inject()(implicit ec: ExecutionContext) extends InboundMessageTransformerInterface {
+
+  def executionContext: ExecutionContext = ec
+
+  override def transform[A](request: ArrivalRequest[A]): Future[InboundRequest[A]] = {
+
+    val inboundMessage: Option[MessageResponse] = request.headers.get("X-Message-Type") match {
+      case Some(MessageType.GoodsReleased.code)             => Some(GoodsReleasedResponse)
+      case Some(MessageType.ArrivalRejection.code)          => Some(ArrivalRejectedResponse)
+      case Some(MessageType.UnloadingPermission.code)       => Some(UnloadingPermissionResponse)
```
amaalali, line 44: Should we move this to the MessageResponse? So something with a signature of `MessageType => MessageResponse`

app/controllers/actions/InboundMessageTransformer.scala
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+  }
+
+}
+
+@ImplementedBy(classOf[InboundMessageTransformer])
+trait InboundMessageTransformerInterface extends ActionTransformer[ArrivalRequest, InboundRequest]
+
+class InboundRequest[A](val inboundMessage: Option[MessageInbound], val request: ArrivalRequest[A]) extends WrappedRequest[A](request)
```
amaalali, line 71: A recent change was to create `models.request` sub-package  where this fits

app/services/SaveMessageService.scala
```diff
@@ -47,7 +47,10 @@ class SaveMessageService @Inject()(arrivalMovementRepository: ArrivalMovementRep
               .addResponseMessage(messageSender.arrivalId, message, arrivalStatus)
               .map {
                 case Success(_) => SubmissionSuccess
-                case Failure(_) => SubmissionFailureInternal
+                case Failure(e) => {
+
+                  SubmissionFailureInternal
+                }
```
amaalali, line 53: Did you want to propagate this back for logging?

test/controllers/actions/InboundMessageTransformerSpec.scala
```diff
@@ -0,0 +1,95 @@
...
+        testAction.futureValue.inboundMessage mustBe
+          Some(MessageInbound(ArrivalRejectedResponse, ArrivalRejected))
+      }
+
+      "IE058" in {
+        val testAction: Future[InboundRequest[_]] = action().transform(fakeRequest("IE058", UnloadingRemarksRejected))
+
+        testAction.futureValue.inboundMessage mustBe
+          Some(MessageInbound(UnloadingRemarksRejectedResponse, UnloadingRemarksRejected))
+      }
```
amaalali, line 90: You could write a generator that is drive off `MessageType.values`

test/controllers/actions/InboundMessageTransformerSpec.scala
```diff
@@ -0,0 +1,95 @@
...
+  val arrivalId        = arrival.arrivalId
+
+  def fakeRequest(code: String, arrivalStatus: ArrivalStatus): ArrivalRequest[AnyContentAsEmpty.type] =
+    ArrivalRequest(FakeRequest("", "").withHeaders("X-Message-Type" -> code), arrival.copy(status = arrivalStatus))
+
+  def action() = new InboundMessageTransformer()
+
+  "InboundMessageTransformer" - {
+
+    "handle unsupported X-Message-Type" in {
```
amaalali, line 55: I think we might be missing the cases where it is a recognised message type, eg `IE007` but should not be accepted by this endpoint?

test/controllers/actions/InboundMessageTransformerSpec.scala
chris-ianson: 👍

test/controllers/actions/InboundMessageTransformerSpec.scala
chris-ianson: These should be covered by refactor to use gen

app/controllers/actions/InboundMessageTransformer.scala
chris-ianson: I've changed the action type to a refiner so it changes the request and also filters within the same action handler (I had added an additional filter action but after thinking it through both actions are closely related so kept them in a single action)

app/controllers/actions/InboundMessageTransformer.scala
chris-ianson: I've put a TODO to update this

app/controllers/actions/InboundMessageTransformer.scala
chris-ianson: Moved

app/services/SaveMessageService.scala
chris-ianson: The logging for failures is handled at controller level. Don't want to change this here because it would mean considering the other failure scenarios (which let the controller handle them)

chris-ianson:
> The `NCTSMessageControllerSpec` is so much cleaner that it really proves the value of the refactor work. In the InboundMessageTransformer we could do a bit more to make the controller body the happy path which is just a little more work

@amaalali Update this based on feedback. There's still a couple of things that could be improved but these are minor I think.

PR #133, opened 2020-07-13
Goods summary

chris-ianson:
Updated ArrivalMessageSummaryService so it returns IE043, IE044, IE058 messages


chris-ianson:
> This looks good 👍
> 
> I think there is probably one more case we want to add a test for (and possibly implement if it doesn't work with the implementation - but I suspect it does)
> 
> As a side note, looking at the model, there are some states that can be represented that we don't want. However the tests arounds the ArrivalMessageSummaryServiceSpec are good and ensure that we don't return those unwanted states. I think thats a good compromise at the moment, and we swap out or representation later if we feel like it

Thats a good point @amaalali I think you're thinking of it from a hateos point and only returning endpoints that are needed (i think?!) If so I'm happy to review this in the future and see what states can be removed for different scenarios (i.e. if they have an IE043 they're never going to need an IO007/IO043?)

PR #139, opened 2020-07-22
CTCTRADERS-717: Backend manage document connector

MatthewAlexanderBeck:


test/connectors/ManageDocumentsConnectorSpec.scala
```diff
@@ -0,0 +1,64 @@
...
+
+  lazy val connector: ManageDocumentsConnector = app.injector.instanceOf[ManageDocumentsConnector]
+
+  implicit val headerCarrier: HeaderCarrier = HeaderCarrier()
+
+  "ManageDocumentsConnector" - {
+
+    "getUnloadingPermissionPdf" - {
+
+      "must return status Ok and PDF" in {
```
amaalali, line 45: Do we use the default exception throwing behaviour of http-verbs? If so can we override it here alone for now (there is a separate tech-debt ticket to fix the others)

test/controllers/PDFGenerationControllerSpec.scala
```diff
@@ -0,0 +1,132 @@
...
+        forAll(genArrival, Arbitrary.arbitrary[ResponseMovementMessage]) {
+          (arrival, responseMovementMessage) =>
+            when(mockArrivalMovementRepository.get(any())).thenReturn(Future.successful(Some(arrival)))
+            when(mockMessageRetrievalService.getUnloadingPermission(any())).thenReturn(Some(responseMovementMessage))
+            when(mockManageDocumentsConnector.getUnloadingPermissionPdf(any())(any())).thenReturn(Future.successful(HttpResponse(200, "")))
+
+            val application = baseApplicationBuilder
+              .overrides(bind[ArrivalMovementRepository].toInstance(mockArrivalMovementRepository))
+              .overrides(bind[MessageRetrievalService].toInstance(mockMessageRetrievalService))
+              .overrides(bind[ManageDocumentsConnector].toInstance(mockManageDocumentsConnector))
```
amaalali, line 76: That's a lot of mocks for a controller, should some of this behaviour be broken out?

test/services/MessageRetrievalServiceSpec.scala
```diff
@@ -0,0 +1,119 @@
...
+
+            val application = baseApplicationBuilder.overrides(bind[ArrivalMessageSummaryService].toInstance(mockArrivalMessageSummaryService)).build()
+
+            val service = application.injector.instanceOf[MessageRetrievalService]
+
+            service.getUnloadingPermission(arrivalWithoutUnloadingPermission) mustBe None
+        }
+      }
+
+      "must return None when UnloadingPermission is in a Failed state" in {
```
amaalali, line 97: This isn't a possible case, Unloading Permission is stateless

PR #142, opened 2020-07-30
CTCTRADERS-718: Updated connector to handle ArrayByte payload

MatthewAlexanderBeck:


app/connectors/ManageDocumentsConnector.scala
```diff
@@ -19,22 +19,24 @@ package connectors
...
     val serviceUrl = s"${config.manageDocumentsUrl}/unloading-permission"
-    val headers    = Seq(("Content-Type", "application/xml"))
+    val headers    = ("Content-Type", "application/xml")
 
-    http.POSTString[HttpResponse](serviceUrl, movementMessage.message.toString, headers)
+    ws.url(serviceUrl)
+      .withHttpHeaders(headers)
+      .post(movementMessage.message.toString)
+      .filter(_.status == 200)
+      .map(_.bodyAsBytes.toArray)
```
amaalali, line 39: So the WSClient throws an exception here, can we prevent that bubbling up and handle it more gracefully?

app/controllers/PDFGenerationController.scala
```diff
@@ -38,23 +38,10 @@ class PDFGenerationController @Inject()(cc: ControllerComponents,
...
-                    logger.error(s"Error when call the PDF Service with the following message: $status - ${request.body}")
-                    InternalServerError
-                  }
-                }
-            }
-            .getOrElse(NotFound)
+        case Some(arrayByte) => Ok(arrayByte)
+        case None =>
+          logger.error(s"Failed to retrieve UnloadingPermission of index: ${request.arrival.arrivalId} ")
+          NotFound
```
amaalali, line 44: We'll also need to have a bad gateway when there is an error with the pdf service, so it returning a 5xx or a timeout. I don't think that this is represented in the return value of UnloadingPermissionPDFService's `getPDF` so we may need to work on that

app/models/WSError.scala
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models
+
+trait WSError
```
amaalali, line 19: Should this be a sealed trait?

PR #148, opened 2020-08-21
CTCTRADERS-721:Audit events for txm

ashwinirudrappa:


app/audit/ArrivalNotificationSubmissionEvent.scala
```diff
@@ -0,0 +1,38 @@
...
+import play.api.libs.json.JsObject
+import play.api.libs.json.Json
+import uk.gov.hmrc.http.HeaderCarrier
+import uk.gov.hmrc.play.audit.AuditExtensions.auditHeaderCarrier
+import uk.gov.hmrc.play.audit.model.ExtendedDataEvent
+import utils.JsonHelper
+
+import scala.xml.NodeSeq
+
+object ArrivalNotificationSubmissionEvent {
```
amaalali, line 28: Do we still need this?

app/utils/JsonHelper.scala
```diff
@@ -0,0 +1,35 @@
...
+
+  def convertXmlToJson(xml: String): JsObject =
+    try {
+      Json.parse(XML.toJSONObject(xml).toString).as[JsObject]
+    } catch {
+      case error: Exception =>
+        Logger.error(s"Failed to convert xml to json with error: ${error.getMessage}")
+        Json.obj() //TODO need to think about it
+    }
+
```
amaalali, line 34: What about changing this to a try here, so 

```
def convertXmlToJson(xml: String): JsObject =
    Try(Json.parse(XML.toJSONObject(xml).toString).as[JsObject])
```

And in the `AuditService` the `auditEvent` method will do the logging, and do a fold/pattern match and fallback to use a string representation if the conversion is a failure. We also do the logging there for diagnostic purposes.

app/utils/JsonHelper.scala
```diff
@@ -0,0 +1,35 @@
...
+import play.api.libs.json.Json
+
+object JsonHelper {
+
+  def convertXmlToJson(xml: String): JsObject =
+    try {
+      Json.parse(XML.toJSONObject(xml).toString).as[JsObject]
+    } catch {
+      case error: Exception =>
+        Logger.error(s"Failed to convert xml to json with error: ${error.getMessage}")
```
amaalali, line 31: What do these error messages look like? We don't want the possibility of the error messages containing any content from the body of the xml

app/audit/AuditService.scala
```diff
@@ -0,0 +1,59 @@
...
+
+class AuditService @Inject()(auditConnector: AuditConnector)(implicit ec: ExecutionContext) {
+
+  def auditEvent(auditType: String, xmlRequestBody: NodeSeq)(implicit hc: HeaderCarrier): Unit = {
+    val eventualJson: Try[JsObject] = JsonHelper.convertXmlToJson(xmlRequestBody.toString())
+    val details = eventualJson match {
+      case Success(data) => data
+      case Failure(error) =>
+        Logger.error(s"Failed to convert xml to json with error: ${error.getMessage}")
+        Json.obj()
```
amaalali, line 44: This stands out to me as something we can probably improve. Lets have a chat with BA about it and see if there is a fallback that we can propose

amaalali (approved):

PR #150, opened 2020-09-10
CTDA-738: Refactor arrivals backend.

marcinkrykowski:


crystalmark (approved):

app/connectors/MessageConnector.scala
```diff
@@ -55,15 +55,13 @@ class MessageConnector @Inject()(config: AppConfig, http: HttpClient)(implicit e
       .copy(authorization = Some(Authorization(s"Bearer ${config.eisBearerToken}")))
       .withExtraHeaders(addHeaders(message.messageType, dateTime, messageSender): _*)
 
+    val possibleResponses           = List(EisSubmissionSuccessful, ErrorInPayload, VirusFoundOrInvalidToken, DownstreamInternalServerError)
+    val statusesOfPossibleResponses = possibleResponses.map(_.httpStatus)
+    val statusToResponseMapping     = statusesOfPossibleResponses.zip(possibleResponses).toMap
```
amaalali, line 60: Seems like a fair refactor, but I'd suggest making this a method on the companion object of `EisSubmissionResult`. So something like 

```
def parseHttpResponse(httpResponse: HttpResponse): EisSubmissionResult
```

marcinkrykowski: Thanks for answering. 
I think in that case we would need to add 4 methods in each class respectively, so we might lose context what such operation is being executed. Also in term on complexity, I think it is better to call map which has effectively constant lookup time than to add more stack frames by calling methods (although it does not have a huge impact). What's more in terms on extendability if we have new response type it's enough to add that type to list instead of adding a new method. What do you think about that Amaal?

marcinkrykowski:
Any more comments on that PR? 😄

jhkldev (approved):

amaalali (approved):

PR #154, opened 2020-10-08
CTDA-803: Return correct status codes in arrivals backend when handling EIS router responses

jhkldev:


app/models/EisSubmissionResult.scala
```diff
@@ -0,0 +1,43 @@
...
+  sealed abstract class EisSubmissionFailure(httpStatus: Int, asString: String) extends EisSubmissionResult(httpStatus, asString)
+
+  sealed abstract class EisSubmissionRejected(httpStatus: Int, asString: String) extends EisSubmissionFailure(httpStatus, asString)
+  object ErrorInPayload                                                          extends EisSubmissionRejected(400, "Message failed schema validation")
+  object VirusFoundOrInvalidToken                                                extends EisSubmissionRejected(403, "Virus found, token invalid etc")
+
+  sealed abstract class EisSubmissionFailureDownstream(httpStatus: Int, asString: String) extends EisSubmissionFailure(httpStatus, asString)
+  object DownstreamInternalServerError                                                    extends EisSubmissionFailureDownstream(500, "Downstream internal server error")
+  case class UnexpectedHttpResponse(httpResponse: HttpResponse)                           extends EisSubmissionFailureDownstream(httpResponse.status, "Unexpected HTTP Response received")
+}
```
amaalali, line 43: Why move these here? This here makes it look like it's generic but it's not since is the contract that the connector provides and contains connector specific concepts such as the http status. It also isn't something that a service (for example) should use as part of it's interface.

test/controllers/MovementsControllerSpec.scala
```diff
@@ -799,8 +885,99 @@ class MovementsControllerSpec extends SpecBase with ScalaCheckPropertyChecks wit
...
+            </CC007A>
+
+          val request = FakeRequest(PUT, routes.MovementsController.putArrival(initializedArrival.arrivalId).url).withXmlBody(requestXmlBody.map(trim))
+
+          val result = route(application, request).value
+
+          status(result) mustEqual INTERNAL_SERVER_ERROR
+
+        }
+      }
```
amaalali, line 978: Why is this an internal server error? This service did not create this payload but is passing along the payload from upstream

app/models/SubmissionProcessingResult.scala
```diff
@@ -16,16 +16,21 @@
...
 
-  sealed trait SubmissionFailure        extends SubmissionProcessingResult
+  sealed trait SubmissionFailure extends SubmissionProcessingResult
+
   case object SubmissionFailureInternal extends SubmissionFailure
+
   case object SubmissionFailureExternal extends SubmissionFailure
-  case object SubmissionFailureRejected extends SubmissionFailure
+
+  case class SubmissionFailureRejected(submissionResult: EisSubmissionRejected) extends SubmissionFailure
```
amaalali, line 33: Returning `EisSubmissionRejected` breaks all encapsulation that this was mean to provide. The controller now uses this for it's implementation

app/controllers/MessagesController.scala
```diff
@@ -68,7 +68,13 @@ class MessagesController @Inject()(
             .map {
               case SubmissionFailureInternal => InternalServerError
               case SubmissionFailureExternal => BadGateway
-              case SubmissionFailureRejected => BadRequest("Failed schema validation")
+              case SubmissionFailureRejected(submissionResult: EisSubmissionRejected) =>
+                submissionResult match {
+                  case ErrorInPayload =>
+                    Status(submissionResult.httpStatus)(submissionResult.asString)
+                  case VirusFoundOrInvalidToken =>
+                    InternalServerError
```
amaalali, line 76: See comment below for `SubmissionFailureRejected `

test/controllers/MessagesControllerSpec.scala
```diff
@@ -371,6 +374,65 @@ class MessagesControllerSpec extends SpecBase with ScalaCheckPropertyChecks with
...
+          .build()
+
+        running(application) {
+          val request = FakeRequest(POST, routes.MessagesController.post(arrival.arrivalId).url).withXmlBody(requestXmlBody)
+
+          val result = route(application, request).value
+
+          status(result) mustEqual INTERNAL_SERVER_ERROR
+        }
+      }
```
amaalali, line 405: See question in MovementsControllerSpec.scala

amaalali (changes requested):

app/models/EisSubmissionResult.scala
jhkldev: I wanted to move this out of the connector as it appears to be generic with just the http status code and message body parameters.

test/controllers/MovementsControllerSpec.scala
jhkldev: The controller recodes the status code to 500 before sending upstream.

app/models/SubmissionProcessingResult.scala
jhkldev: I could change EisSubmissionRejected so that it implements the SubmissionFailure trait but that would mean removing the sealed trait.

test/controllers/MovementsControllerSpec.scala
amaalali: After discussion we think that the spec is not granular enough and downstream from the router should instead return a 401 if the auth tokens are missing/invalid

If we cannot resolve this will be good to go

test/controllers/MessagesControllerSpec.scala
amaalali: See https://github.com/hmrc/transit-movements-trader-at-destination/pull/154#discussion_r505456552

test/controllers/MovementsControllerSpec.scala
amaalali: Looking over the spec this is a misunderstanding on my part, this is good to the spec

app/models/SubmissionProcessingResult.scala
```diff
@@ -16,16 +16,34 @@
...
+  case class SubmissionFailureRejected(httpStatus: Int, asString: String) extends SubmissionFailure {
+
+    def recodeForUpstream(): Result =
+      httpStatus match {
+        case BAD_REQUEST =>
+          BadRequest(asString)
+        case FORBIDDEN =>
+          InternalServerError
+      }
+  }
```
amaalali, line 46: There is no need for the recode or expose these details outside of the service layer or need . The Forbidden shouldn't be mapped to this at all, this indicates an error in logic somewhere else

amaalali (changes requested):

amaalali (approved):

PR #158, opened 2020-11-18
CTDA-820: Remove raw XML from AuditDetails

marcinkrykowski:


RobertLaverick (changes requested):
This PR needs to wait for some other changes to be merged in, but it will mean the model still doesn't match the TXM document. lets try and accomodate that in this PR.

crystalmark:
> This PR needs to wait for some other changes to be merged in, but it will mean the model still doesn't match the TXM document. lets try and accomodate that in this PR.

Do we need to change the ticket so instead of "remove xml" it is more like "change arrivals audit events to match departures document structure"?  Or should it be a separate ticket to do the rest of the changes?

RobertLaverick:
> > This PR needs to wait for some other changes to be merged in, but it will mean the model still doesn't match the TXM document. lets try and accomodate that in this PR.
> 
> Do we need to change the ticket so instead of "remove xml" it is more like "change arrivals audit events to match departures document structure"? Or should it be a separate ticket to do the rest of the changes?

I mean, we may as well change the ticket. Either works, though I have just noticed that its going to be true for both services - lets have this one focus on the Arrivals backend.

It's also got a link to our TxM docs rather than the Arrivals TxM docs, which whilst not a problem since the model is the same, is a little embarrassing. I'll confirm the model and update the ticket with a new link if we're going with "CTDA-820 is finishing up the arrivals audit model"

RobertLaverick:
> > > This PR needs to wait for some other changes to be merged in, but it will mean the model still doesn't match the TXM document. lets try and accomodate that in this PR.
> > 
> > 
> > Do we need to change the ticket so instead of "remove xml" it is more like "change arrivals audit events to match departures document structure"? Or should it be a separate ticket to do the rest of the changes?
> 
> I mean, we may as well change the ticket. Either works, though I have just noticed that its going to be true for both services - lets have this one focus on the Arrivals backend.
> 
> It's also got a link to our TxM docs rather than the Arrivals TxM docs, which whilst not a problem since the model is the same, is a little embarrassing. I'll confirm the model and update the ticket with a new link if we're going with "CTDA-820 is finishing up the arrivals audit model"

I've confirmed with @amaalali that he's happy with use using our schema for their events at the moment since their TxM document doesn't have any examples at this moment. So we can use that as a guide.

app/audit/AuditDetails.scala
```diff
@@ -16,12 +16,14 @@
...
-case class AuditDetails(json: JsObject, xml: String)
+case class AuditDetails(json: JsObject)
 
 object AuditDetails {
-  implicit val writes: OWrites[AuditDetails] = Json.writes[AuditDetails]
+  implicit val writes: OWrites[AuditDetails] = new OWrites[AuditDetails] {
+
+    def writes(details: AuditDetails): JsObject =
+      JsObject(details.json.fields.toMap)
+  }
```
RobertLaverick, line 28: We need tests for this, can you please add tests for this behaviour

RobertLaverick (changes requested):

app/audit/AuditDetails.scala
marcinkrykowski: 👍

app/audit/AuditDetails.scala
```diff
@@ -16,12 +16,14 @@
...
 
 case class AuditDetails(json: JsObject)
 
 object AuditDetails {
-  implicit val writes: OWrites[AuditDetails] = Json.writes[AuditDetails]
+  implicit val writes: OWrites[AuditDetails] = new OWrites[AuditDetails] {
+
+    def writes(details: AuditDetails): JsObject =
+      JsObject(details.json.fields.toMap)
+  }
```
amaalali, line 28: I don't think we need the intermediate JsObject

```suggestion
  implicit val writes: OWrites[AuditDetails] = new OWrites[AuditDetails] {

    def writes(details: AuditDetails): JsObject = details.json
  }
```

app/audit/AuditDetails.scala
```diff
@@ -16,12 +16,14 @@
 
 package audit
 
-import play.api.libs.json.JsObject
-import play.api.libs.json.Json
-import play.api.libs.json.OWrites
+import play.api.libs.json._
 
-case class AuditDetails(json: JsObject, xml: String)
+case class AuditDetails(json: JsObject)
```
amaalali, line 21: Do we even need this class anymore?

amaalali (commented):
One main comment that may simplify things. Do we even need the AuditDetails class anymore?

This work doesn't add the channel information? Is that a separate piece of work?

app/audit/AuditDetails.scala
RobertLaverick: If you're going to use the same model as us, then I expect this will include the channel field.

marcinkrykowski:
If arrivals gonna match the departure audit model than we should keep `AuditDetails` class as we might need that when adding mentioned `channel`, but that's a different piece of work. If you guys @amaalali @RobertLaverick are happy with that PR let's merge it! :D

RobertLaverick (approved):

amaalali (approved):

PR #159, opened 2020-11-23
added webDefault channel headers to audit events

RobertLaverick:


app/models/request/ChannelOptionalRequest.scala
```diff
@@ -0,0 +1,27 @@
...
+package models.request
+
+import play.api.mvc.WrappedRequest
+
+trait ChannelOptionalRequest[A] extends WrappedRequest[A] {
+
+  def getChannel: String = headers.get("channel") match {
+    case Some(channel) => channel
+    case None          => "web"
+  }
```
amaalali, line 26: I don't think we should have any default, if the header isn't specified then it should be a Bad Request.

Since the FE hasn't implemented the work to incorporate with this, this is a buildup of technical debt.

In the mean time, while we live with the TD, this isn't adding an _optional_ field of channel, so I don't think that this name is correct

amaalali (changes requested):

app/controllers/MovementsController.scala
```diff
@@ -136,8 +136,8 @@ class MovementsController @Inject()(
               case submissionFailureRejected: SubmissionFailureRejected =>
                 BadRequest(submissionFailureRejected.responseBody)
               case SubmissionSuccess =>
-                auditService.auditEvent(AuditType.ArrivalNotificationReSubmitted, request.body)
-                auditService.auditEvent(AuditType.MesSenMES3Added, message.message)
+                auditService.auditEvent(AuditType.ArrivalNotificationReSubmitted, request.body, request.getChannel)
+                auditService.auditEvent(AuditType.MesSenMES3Added, message.message, request.getChannel)
                 Accepted("Message accepted")
                   .withHeaders("Location" -> routes.MovementsController.getArrival(request.arrival.arrivalId).url)
             }
```
GeorgeColdham, line 143: Possibly not for this PR, but as far as I can tell this map is repeated multiple times and is the same code block. could this functionality be moved to a single function?

marcinkrykowski (approved):

GeorgeColdham (approved):
Looks fine to me, my only comment is what I left before. I am possibly not the only set of eyes you want approving this though :p

app/controllers/MovementsController.scala
RobertLaverick: It took me a second to see what you meant here because of the line selection, but that's a fair point for a refactor. @GeorgeColdham would you write it up as a low-priority tech-debt ticket? Or potentially a low-priority spike?

jhkldev (approved):

app/controllers/MovementsController.scala
GeorgeColdham: https://jira.tools.tax.service.gov.uk/browse/CTDA-961

app/models/ChannelType.scala
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package models
+
+sealed trait ChannelType
+
+object ChannelType extends Enumerable.Implicits {
+  case object web  extends ChannelType
+  case object api  extends ChannelType
+  case object ncts extends ChannelType
```
amaalali, line 24: We need to confirm that we need to add "ncts" for the channel. Should this not only be web/api?

amaalali (changes requested):

app/models/ChannelType.scala
RobertLaverick: at the moment its because audit-wise its behaving as an origin-tracker, but I appreciate that in the departure model the channel will be used for filtering. So when that works happens ncts may well be removed and instead source the channel of the departure directly from that route. But at the moment there is no way to discern the correct channel.

RobertLaverick: @amaalali I've spoken with Maria Decosta about this and she's of the opinion that ncts doesn't add value (which I agree with). I was going to remove this as part of the work to only provide web or api departures/arrivals to web/api, are you happy with that or would you rather I remove nct and put in a different default value for now?

amaalali (approved):

PR #160, opened 2020-11-24
Update header for x-forwarded-host to be MDTP

amaalali:


PSouthern:
@amaalali what about the `messageSender` model?  Should that also be changed to only accept MDTP, it currently allows mixed case.

PR #164, opened 2020-12-03
CTCTRADERS-1319: Record metrics on messages received from NCTS

TheNickWilson:


app/controllers/NCTSMessageController.scala
```diff
@@ -47,11 +50,17 @@ class NCTSMessageController @Inject()(cc: ControllerComponents,
...
 
       processingResult map {
-        case SubmissionSuccess =>
-          auditService.auditNCTSMessages(messageInbound.messageType, xml)
-          Ok
-        case SubmissionFailureInternal => internalServerError("Internal Submission Failure " + processingResult)
-        case SubmissionFailureExternal => badRequestError("External Submission Failure " + processingResult)
+        result =>
+          val counter = Monitors.countMessagesFromNcts(messageInbound.messageType.messageType, request.request.getChannel, result)
+          metricsService.inc(counter)
```
amaalali, line 55: So I'm not sure where this should be so I'm opening it as a conversation. Should we have this here, so always, or should we only increment the counter when we return `Ok`?

app/metrics/Monitors.scala
```diff
@@ -61,4 +64,7 @@ object Monitors {
       case s if s <= 500 => Counter("arrivals-per-eori-251-500")
       case _             => Counter("arrivals-per-eori-501-or-more")
     }
+
+  def countMessagesFromNcts(messageType: MessageType, channel: ChannelType, outcome: SubmissionProcessingResult): Counter =
```
amaalali, line 68: This can be used more generically, possibly on the out bound side. For example we can count number of IE015 vs IE014 (assuming that's a useful metric).

So this could just be (in words) _count the number of messages by the message type_

amaalali (commented):
One small questions about where we implement the metrics for the NCTS messages.

Also, should we collect metrics that we send down to NCTS? Knowing what fraction that need to be cancelled might be useful information

app/controllers/NCTSMessageController.scala
TheNickWilson: I've included the result of processing the message in the counter, so we can distinguish between success or failure (e.g. we'll have separate counters for "web, IExxx, success", "web, IExxx, internal error" etc.). I thought this would be useful, but can strip it out and only record successes if you prefer.

app/metrics/Monitors.scala
TheNickWilson: Good point!  I'll take a look at whether this can be applied to the outbound stuff too.  You're right that it might be useful to graph them all together.

PR #167, opened 2020-12-10
CTDA-994 - Convert field names to human-readable form for audit events

TheNickWilson:


test/utils/MessageTranslationSpec.scala
```diff
@@ -0,0 +1,44 @@
...
+      val app = appBuilder.build()
+
+      running(app) {
+        val service = app.injector.instanceOf[MessageTranslation]
+        val input   = "field1, field1, field2, field3"
+        val result  = service.translate(input)
+
+        result mustEqual "Description 1, Description 1, Description 2, field3"
+      }
+    }
```
amaalali, line 42: Do you mind splitting out this test into 2 scenarios, 1 where all the keys are known and the other where there is an unknown field. I see that it's documented in the JsonHelperSpec but I think that behaviour lives here

chris-ianson (approved):

amaalali (changes requested):
More than happy to be disagreed with about this change

TheNickWilson:
Test this please

amaalali (approved):

PR #181, opened 2021-01-11
Ctctraders 1707-  return BAD_GATEWAY if downstream fails

kenneth-rayner:


amaalali (approved):
Looks good. Since this changes the interface of the API, we need to change any consuming services

PR #185, opened 2021-01-15
CTDA-961/Movements Controller repeating code cleanup

GeorgeColdham:
# Movements Controller Cleanup

cleanup

[Ticket](https://jira.tools.tax.service.gov.uk/browse/CTDA-961)

### DIFF
--- a/app/controllers/MovementsController.scala
+++ b/app/controllers/MovementsController.scala
@@ -72,7 +72,7 @@ class MovementsController @Inject()(
       case _                                                           => false
     }
 
-  private def SubmissionResultHandler(
+  private def handleSubmissionResult(
     result: SubmissionProcessingResult,
     arrivalNotificationType: String,
     message: NodeSeq,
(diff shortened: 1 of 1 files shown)


app/controllers/MovementsController.scala
```diff
@@ -68,6 +72,25 @@ class MovementsController @Inject()(
       case _                                                           => false
     }
 
+  private def SubmissionResultHandler(
```
RobertLaverick, line 75: typically defs start lowercase

RobertLaverick: Either lowercase this, or if we want to be extra clever, check the controller tests to see if they track the auditing - if they don't, lets move this into its own class that can be instantiated as this is used, but has its own suite of unit tests to check the audits, etc.

RobertLaverick (changes requested):

app/controllers/MovementsController.scala
TheNickWilson: I'd avoid moving this to a separate class - it's very much the concern of the controller to decide what to return.  Are we also 100% confident the different methods will never want to handle the result differently in the future?  DRY's all well and good, but you're also coupling the behaviour of 2 distinct methods - just worth a sanity check to make sure that's valid!  If we're happy they really ought to be the same, then crack on! (with the lowercasing Rob suggests - and maybe using a more active name, e.g. handleSubmissionResult?)

RobertLaverick (approved):

PR #187, opened 2021-01-22
CTCTRADERS-1811: Worker to add JSON to messages

TheNickWilson:
This PR adds a worker that will add the JSON representation of message payloads to records that don't have them.

### DIFF
--- a/app/models/LockResult.scala
+++ b/app/models/LockResult.scala
@@ -1,0 +1,24 @@
+/*
+ * Copyright 2021 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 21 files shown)


amaalali (approved):
There are some things here worth abstracting so that we have a generalised mechanism for migrations and some decomposition of some of the worker, in particular `AddJsonToMessagesWorker`, but this looks good for the specific purpose of this data migration.

TheNickWilson:
@amaalali most of the worker could be abstracted, leaving the implementations to basically just specify which lock they require, what their settings are, and what the actual work entails.

I kept everything together since this is a temporary worker that'll get deleted (and I'd envisioned the entire workers package would get removed).  If you'd prefer that we do the abstraction and have all of the "plumbing" kept there, so workers are easier to write in future, I can do that.

amaalali:
That's not a requirement for this change, so it would be unreasonable to insist on that now. But when we need to remove it, I think there is so much of value here that we can disable and then strip back the bits specific to this migration and leave a general migration framework in place

PR #190, opened 2021-01-27
[CTCTRADERS-1859] Pull less data back for arrivals

ericohwotu:
Separate out the mongo local date reads from plays local date reads to prevent invalid Json being passed to the frontend


ericohwotu:
test this please

amaalali (approved):

PR #192, opened 2021-02-03
Ctctraders 1906 add 917 message response

kenneth-rayner:


amaalali (approved):

amaalali:
The API Postmans tests for this service needs to be changed for this PR also. See similar PR https://github.com/hmrc/transit-movements-trader-postman/pull/53

ashwinirudrappa:
test this please

app/models/ArrivalStatus.scala
```diff
@@ -67,6 +69,13 @@ object ArrivalStatus extends Enumerable.Implicits with MongoDateTimeFormats {
     }
   }
 
+  case object XMLSubmissionNegativeAcknowledgement extends ArrivalStatus {
+    override def transition(messageReceived: MessageReceivedEvent): Either[TransitionError, ArrivalStatus] = messageReceived match {
+      case MessageReceivedEvent.XMLSubmissionNegativeAcknowledgement => Right(XMLSubmissionNegativeAcknowledgement)
+      case _                                                         => Left(TransitionError(s"Tried to transition from XMLSubmissionNegativeAcknowledgement to $messageReceived."))
```
crystalmark, line 75: Are we not allowing any more messages for a movement after a 917?

kenneth-rayner: As of now it is 917 response based on the notifications we have received for live issues

RobertLaverick (changes requested):
XMLSubmissionNegativeAcknowledgement is a bit confusing as names go since it suggests that there is a schema/xml-level problem, when to my knowledge 917 is a content-issue.

app/models/ArrivalStatus.scala
```diff
@@ -67,6 +69,13 @@ object ArrivalStatus extends Enumerable.Implicits with MongoDateTimeFormats {
     }
   }
 
+  case object XMLSubmissionNegativeAcknowledgement extends ArrivalStatus {
```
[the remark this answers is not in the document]
ashwinirudrappa: @RobertLaverick This is the actual definition of IE917,we wanted to keep it that way so we know what we are dealing with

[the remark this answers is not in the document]
RobertLaverick: I appreciate that, but this is also a field that will be returned to the user if they want to check the arrival status, we should be concerned whether this is meaningful to them, or confusing. Further, I'm pretty sure none of the other status's match verbatim.

Let me have a glance at the DDNTA and I'll comment again then.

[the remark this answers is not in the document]
ashwinirudrappa: @RobertLaverick We do not show this status to Users, it is only for us to know on the UI they still see this as "arrival rejection"

app/models/ArrivalStatus.scala
RobertLaverick: This does raise alarm bells - do we have any tickets or intention to follow this up? We've effectively ended an arrival permanently with this transition and forced a new one, it might even be worse because if you try to resubmit with the same MRN, it will try to merge, transition and kill that arrival too.

app/models/ArrivalStatus.scala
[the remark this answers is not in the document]
RobertLaverick: This will be returned on the API-side rather than the Web frontend.

Regardless, I'm fine with this naming after a chat with @crystalmark

app/models/request/actions/InboundMessageTransformer.scala
```diff
@@ -67,12 +69,15 @@ class InboundMessageTransformer @Inject()(implicit ec: ExecutionContext) extends
...
-  private[models] def messageResponse(code: Option[String]): Option[MessageResponse] = code match {
+  private[models] def messageResponse(code: Option[String], channel: ChannelType): Option[MessageResponse] = code match {
     case Some(MessageType.GoodsReleased.code)             => Some(GoodsReleasedResponse)
     case Some(MessageType.ArrivalRejection.code)          => Some(ArrivalRejectedResponse)
     case Some(MessageType.UnloadingPermission.code)       => Some(UnloadingPermissionResponse)
     case Some(MessageType.UnloadingRemarksRejection.code) => Some(UnloadingRemarksRejectedResponse)
-    case _                                                => None
+    case Some(MessageType.XMLSubmissionNegativeAcknowledgement.code) =>
+      logger.error(s"Received the message ${MessageType.XMLSubmissionNegativeAcknowledgement.code} for the $channel channel")
+      Some(XMLSubmissionNegativeAcknowledgementResponse)
```
RobertLaverick, line 79: I appreciate this is logging at an extreme level, but is there any place in the incoming journey that will audit the 917 so we can at least look at an obfuscated 917 to understand the errors?

ashwinirudrappa: There is no separate auditing for 917 it follows the same structure that we have for all incoming messages

ashwinirudrappa: We have configured an alert for this message especially

RobertLaverick: we won't be able to diagnose the cause of the problem with a simple alert though. Sure we'll know it has happened, but not WHY. That's the value of an audit.

RobertLaverick: Pagerduty alerts tell us it has happened, not what the core issue is. We need some way of identifying what the actual problem is with the XML. We can’t just look it up in the database in production.

RobertLaverick: I'd overlooked that we already audit incoming events. Issue resolved.

app/models/ArrivalStatus.scala
RobertLaverick: Also, if this is going to be included, then we should confirm this behaviour in an acceptance test.

amaalali: This is parity with the current system, if there is an IE917 then that's the end of the movement and then the submitting system needs to raise a new movement. It would be nice to group it together on our end for APIs / Frontends to consume but that hasn't been done as yet. 

This is also in effect the same as the current process for an IE008 with the FE where for some error types we don't group together under the same movement. 

This isn't well known and understood this branch of errors and what the followup should be so this is open to change as we learn more about the process after receipt of an IE917

crystalmark (approved):

PR #196, opened 2021-02-10
Made one of the test dates static to remove the race condition

crystalmark:
CTDA-1124 Made one of the test dates static to remove the race condition when milliseconds are 0.

### DIFF
--- a/test/controllers/MovementsControllerSpec.scala
+++ b/test/controllers/MovementsControllerSpec.scala
@@ -75,7 +75,7 @@ class MovementsControllerSpec extends SpecBase with ScalaCheckPropertyChecks wit
   val localTime     = LocalTime.of(1, 1)
   val localDateTime = LocalDateTime.of(localDate, localTime)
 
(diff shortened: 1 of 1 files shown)


test/controllers/MovementsControllerSpec.scala
```diff
@@ -73,6 +75,8 @@ class MovementsControllerSpec extends SpecBase with ScalaCheckPropertyChecks wit
   val localTime     = LocalTime.of(1, 1)
   val localDateTime = LocalDateTime.of(localDate, localTime)
 
+  val dateFormat = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss'.'SS")
```
amaalali, line 78: Should we use this formatter explicitly in ResponseArrival? Also I think we can reference: `DateTimeFormatter.ISO_LOCAL_DATE_TIME`

amaalali (approved):

kenneth-rayner (approved):

PR #198, opened 2021-02-12
CTCTRADERS 1811 refactor worker

amaalali:
- Move logic for adding json transformation of xml message to class.
- Move locking into it's own class
- Allow the repository to stream the arrivals with messages that don't have Json
- Add flow that produces Arrivals when the stream has a successful lock
- Add flow for transformer
- Refactor AddJsonToMessagesWorker to use smaller flows in workers package.


amaalali:
test this please

TheNickWilson (approved):

PR #202, opened 2021-02-22
Rewrite test for InboundMessageTransformer.

amaalali:


test/models/request/actions/InboundMessageTransformerSpec.scala
```diff
@@ -54,78 +50,102 @@ import scala.concurrent.Future
...
   "InboundMessageTransformer" - {
 
-    "handle unsupported `X-Message-Type`" in {
+    "returns 200 for supported `X-Message-Type`" in {
+      forAll(arbitrary[ChannelType]) {
+        channel =>
+          val arrivalMovement = arrival.copy(status = ArrivalStatus.ArrivalSubmitted)
+
+          val request =
+            ArrivalRequest(FakeRequest("", "").withHeaders("X-Message-Type" -> MessageType.ArrivalRejection.code), arrivalMovement, channel)
```
MatthewAlexanderBeck, line 74: Do we maybe want to generate the MessageType to make this test a bit more robust (or is that overkill?)

MatthewAlexanderBeck (approved):

app/models/MessageType.scala
```diff
@@ -39,10 +39,10 @@ object MessageType extends Enumerable.Implicits {
     Seq(
       ArrivalNotification,
       ArrivalRejection,
-      GoodsReleased,
       UnloadingPermission,
       UnloadingRemarks,
       UnloadingRemarksRejection,
+      GoodsReleased,
```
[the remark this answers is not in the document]
amaalali: No, not any impact that's picked up by the unit or integration tests

test/models/request/actions/InboundMessageTransformerSpec.scala
amaalali: I don't think it is possible :( since there is validation on the X-Message-Type being a known value AND it must be a value that we know how to handle given all the messages we've see so far. Any generators to make this possible are likely to push the code into the unintelligible region

PR #209, opened 2021-03-03
CTCTRADERS-1811 Improve supervision configuration in workers

amaalali:


app/workers/AddJsonToMessagesTransformer.scala
```diff
@@ -45,38 +46,15 @@ private[workers] class AddJsonToMessagesTransformer @Inject()(
...
-
-    case NonFatal(e) =>
-      logger.warn("Worker saw an exception but will resume", e)
-      Supervision.resume
-
-    case e =>
-      logger.error("Worker saw a fatal exception and will be stopped", e)
-      Supervision.stop
-  }
+  private val supervisionStrategy: WorkerProcessingProblemSupervisionStrategyProvider =
```
TheNickWilson, line 49: Would changing the name of this val as well let us avoid having `supervisionStrategy.supervisionStrategy` later?

TheNickWilson (approved):

app/workers/AddJsonToMessagesTransformer.scala
amaalali: Done. I've renamed the `WorkerProcessingProblemSupervisionStrategyProvider` to `WorkerSupervisionDeciderProvider` and the variable names to `supervisionDeciderProvider`.

PR #211, opened 2021-03-24
[CLI-46] prevent unloading remarks from being submitted multiple times without a response

ericohwotu:


app/controllers/actions/MessageTransformer.scala
```diff
@@ -0,0 +1,88 @@
...
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+import scala.xml.NodeSeq
+
+class MessageTransformer @Inject()(implicit val executionContext: ExecutionContext) extends MessageTransformerInterface with Logging {
+
+  override def refine[A](request: ArrivalRequest[A]): Future[Either[Result, MessageTransformRequest[A]]] =
+    request.body match {
+      case x: NodeSeq =>
+        x.headOption.flatMap(node => messageResponse(request.channel)(node.label)) match {
```
MatthewAlexanderBeck, line 46: Might be worth double checking that the head node will always be the identifier. Not sure if api adds any wrappers etc.

app/models/ArrivalStatus.scala
```diff
@@ -57,6 +56,7 @@ object ArrivalStatus extends Enumerable.Implicits with MongoDateTimeFormats {
     }
   }
 
+  /*TODO: investigate this as part of a separate ticket*/
```
MatthewAlexanderBeck, line 59: Can we remove this when the ticket is created as this is live code

test/controllers/actions/MessageTransformerSpec.scala
```diff
@@ -33,63 +30,84 @@ import models.UnloadingRemarksRejectedResponse
...
 
           val request =
-            ArrivalRequest(FakeRequest("", "").withHeaders("X-Message-Type" -> MessageType.ArrivalRejection.code), arrivalMovement, channel)
+            ArrivalRequest(
+              FakeRequest("", "")
+                .withHeaders(
+                  "X-Message-Type" -> MessageType.ArrivalRejection.code,
+                  "Content-Type"   -> "application/xml"
+                )
+                .withBody[NodeSeq](<CC008A> </CC008A>),
```
MatthewAlexanderBeck, line 85: Is it worth adding a Gen here for the rootnodes to test them all (we have them available in the MessageType model)

app/models/ArrivalStatus.scala
ericohwotu: Removing

test/controllers/actions/MessageTransformerSpec.scala
ericohwotu: This also depends on the status of the arrival so might be a complex generator. will test all of them manually

MatthewAlexanderBeck (approved):

crystalmark (approved):

PR #214, opened 2021-03-31
CTCTRADERS-2084: Seed multiple arrivals

MatthewAlexanderBeck:


app/testOnly/services/TestDataXMLGenerator.scala
```diff
@@ -0,0 +1,97 @@
...
+
+    val xml = TestDataXMLGenerator.arrivalNotification(mrn.format, eori.format)
+
+    val movementMessage = MovementMessageWithStatus(dateTime, ArrivalNotification, xml, MessageStatus.SubmissionPending, 1)
+
+    Arrival(arrivalId,
+            ChannelType.web,
+            MovementReferenceNumber(mrn.format),
+            eori.format,
+            Initialized,
```
ericohwotu, line 51: When the data is inserted will there be a way to transition out of Initialised? Do we plan on using the test-support service with these users?

amaalali: > When the data is inserted will there be a way to transition out of Initialised?

That's a great spot Eric! No there is nothing that will interact with these to transition the state from Initialized after this data is seeded. I'll update this to be ArrivalSubmitted

> Do we plan on using the test-support service with these users?

No plans currently but we can use it on an ad-hoc basis going forward, which means that the status needs to be ArrivalSubmitted as you pointed out above

I'll get this fixed and updated asap

app/testOnly/services/TestDataGenerator.scala
```diff
@@ -0,0 +1,99 @@
...
+
+private[services] class TestDataGenerator @Inject()(clock: Clock) {
+
+  def arrivalMovement(eori: SeedEori, mrn: SeedMrn, arrivalId: ArrivalId): Arrival = {
+
+    val dateTime = LocalDateTime.now(clock)
+
+    val xml = TestDataXMLGenerator.arrivalNotification(mrn.format, eori.format)
+
+    val movementMessage = MovementMessageWithStatus(dateTime, ArrivalNotification, xml, MessageStatus.SubmissionPending, 1)
```
ericohwotu, line 45: Last thing i think status should be submitted

ericohwotu:
I think we also need an prod config change to disable the feature flag

ericohwotu (approved):

PR #218, opened 2021-04-30
CTDA-1308 Filter arrival messages by last updated timestamp

DavidGregory084:
This adds an optional `updatedSince` query parameter to the `getAllArrivals` endpoint. This will allow API users to cut down the amount of data they need to fetch at one time.

I have also included the changes required to use the new artifact store due to the Bintray decommissioning.

### DIFF
--- a/app/models/ArrivalUpdate.scala
+++ b/app/models/ArrivalUpdate.scala
@@ -68,7 +68,7 @@ object MessageStatusUpdate extends MongoDateTimeFormats {
           "$set" ->
             Json.obj(
               s"messages.${value.messageId.index}.status" -> value.messageStatus,
-              "lastUpdated"                               -> LocalDateTime.now(clock).withSecond(0).withNano(0)
+              "lastUpdated"                               -> LocalDateTime.now(clock)
             )
       )
     )
@@ -83,7 +83,7 @@ object ArrivalStatusUpdate extends MongoDateTimeFormats {
       Json.obj(
         "$set" -> Json.obj(
           "status"      -> value.arrivalStatus,
-          "lastUpdated" -> LocalDateTime.now(clock).withSecond(0).withNano(0)
+          "lastUpdated" -> LocalDateTime.now(clock)
         )
     )
 }
@@ -105,7 +105,7 @@ object ArrivalPutUpdate extends MongoDateTimeFormats {
       Json.obj(
         "$set" -> Json.obj(
           "movementReferenceNumber" -> a.movementReferenceNumber,
-          "lastUpdated"             -> LocalDateTime.now(clock).withSecond(0).withNano(0)
+          "lastUpdated"             -> LocalDateTime.now(clock)
         )
       ) deepMerge ArrivalModifier.toJson(a.arrivalUpdate)
 
--- a/app/models/ArrivalUpdate.scala
+++ b/app/models/ArrivalUpdate.scala
@@ -16,6 +16,7 @@
 
 package models
 
+import java.time.Clock
 import java.time.LocalDateTime
 
 import cats._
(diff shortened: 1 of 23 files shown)


app/models/ArrivalUpdate.scala
```diff
@@ -59,14 +60,15 @@ object ArrivalUpdate {
...
+
+  implicit def arrivalStateUpdate(implicit clock: Clock, writes: Writes[MessageStatus]): ArrivalModifier[MessageStatusUpdate] =
     ArrivalModifier(
       value =>
         Json.obj(
           "$set" ->
             Json.obj(
               s"messages.${value.messageId.index}.status" -> value.messageStatus,
-              "lastUpdated"                               -> LocalDateTime.now.withSecond(0).withNano(0)
+              "lastUpdated"                               -> LocalDateTime.now(clock).withSecond(0).withNano(0)
```
DavidGregory084, line 71: This use of `LocalDateTime.now` was causing nondeterministic failures in the test suite.

I'm not wild about plumbing a `Clock` through everything implicitly but there's little other choice

build.sbt
```diff
@@ -1,75 +1,97 @@
...
 val appName = "transit-movements-trader-at-destination"
 
 lazy val microservice = Project(appName, file("."))
   .enablePlugins(play.sbt.PlayScala, SbtAutoBuildPlugin, SbtGitVersioning, SbtDistributablesPlugin, SbtArtifactory)
-  .disablePlugins(JUnitXmlReportPlugin)
+  .disablePlugins(JUnitXmlReportPlugin) //Required to prevent https://github.com/scalatest/scalatest/issues/1427
   .configs(IntegrationTest)
-  .settings(inConfig(IntegrationTest)(itSettings): _*)
-  .settings(inConfig(IntegrationTest)(scalafmtSettings): _*)
+  .settings(DefaultBuildSettings.integrationTestSettings())
```
DavidGregory084, line 11: Most of the changes here relate to using default settings from the sbt-settings plugin rather than rolling our own

it/api/OutboundMessagesApiSpec.scala
```diff
@@ -79,30 +96,122 @@ class OutboundMessagesApiSpec extends AnyFreeSpec with GuiceOneServerPerSuite wi
         </HEAHEA>
       </CC044A>
 
-      repo.insert(arrival).futureValue
+      await(repo.insert(arrival))
```
DavidGregory084, line 99: `.futureValue` does not use the default await timeout from the `FutureAwaits` helper so these tests were flaky

project/plugins.sbt
```diff
@@ -1,21 +1,21 @@
-resolvers += Resolver.url("HMRC Sbt Plugin Releases", url("https://dl.bintray.com/hmrc/sbt-plugin-releases"))(
-  Resolver.ivyStylePatterns)
+resolvers += "HMRC-open-artefacts-maven" at "https://open.artefacts.tax.service.gov.uk/maven2"
 
-resolvers += "HMRC Releases" at "https://dl.bintray.com/hmrc/releases"
+resolvers += Resolver.url("HMRC-open-artefacts-ivy", url("https://open.artefacts.tax.service.gov.uk/ivy2"))(Resolver.ivyStylePatterns)
 
-resolvers += Resolver.typesafeRepo("releases")
+addSbtPlugin("uk.gov.hmrc" % "sbt-auto-build" % "2.15.0")
```
DavidGregory084, line 5: I've included the Bintray resolvers changes in here

test/models/ArrivalStatusSpec.scala
```diff
@@ -28,23 +28,23 @@ class ArrivalStatusSpec extends SpecBase with ScalaCheckDrivenPropertyChecks wit
     "Initialized must" - {
 
       "transition to ArrivalSubmitted when receiving a ArrivalSubmitted event" in {
-        Initialized.transition(MessageReceivedEvent.ArrivalSubmitted).right.value mustBe ArrivalSubmitted
+        Initialized.transition(MessageReceivedEvent.ArrivalSubmitted).value mustBe ArrivalSubmitted
```
DavidGregory084, line 31: `Either` is right-biased as of Scala 2.12+ so the right projection isn't needed

app/models/ArrivalUpdate.scala
amaalali: I think the withSecond and withNano can be removed now and an appropriate clock can be used in the tests

DavidGregory084:
test this please

amaalali (approved):

amaalali (commented):
Reflecting on the need to implicitly use clock for the lastUpdated field, this is only forced because we are adding that field at the JSON transformation stage.

One way to eliminate this implicit clock (which has become parasitic) is to move the time stamp to being a value on the `ArrivalUpdate`. This would mean the clock would only be needed where we create the update class. This would allow all us to eliminate the need for a clock in the repository. 

This is a bit of refactor so we can still decide on how to approach this.

PR #222, opened 2021-05-18
Fix date filtering implementation and tests

DavidGregory084:
I think the date filtering selector was lost to a merge conflict. I have restored that now! 

Unfortunately the test for this was still passing due to the way that the capping was implemented - the last 2 results were returned in reverse updated order due to the capping config for api.maxRowsReturned being 2, which had the same effect as the date filter in the integration tests. I've updated the test to use the web channel.


RobertLaverick (approved):

DavidGregory084:
test this please

crystalmark (approved):

DavidGregory084:
test this please

PR #224, opened 2021-05-26
Use X-Client-Id header rather than auth Retrievals

DavidGregory084:
This makes this functionality a lot easier to test locally because we can simply pass this header rather than using auth Retrievals. In production this header is added by the API Platform.


DavidGregory084:
rebuild

RobertLaverick (approved):

PR #233, opened 2021-07-08
modified push notification model to replace body with messageUri

RobertLaverick:


crystalmark (approved):

app/models/ArrivalMessageNotification.scala
```diff
@@ -53,12 +55,16 @@ object ArrivalMessageNotification {
         obj ++ Json.obj("requestId" -> requestId(arrival.arrivalId))
     }
 
-  def fromRequest(request: InboundMessageRequest[NodeSeq], timestamp: LocalDateTime): ArrivalMessageNotification =
+  def fromRequest(request: InboundMessageRequest[NodeSeq], timestamp: LocalDateTime): ArrivalMessageNotification = {
+    val messageId  = MessageId.fromIndex(request.arrivalRequest.arrival.messages.length)
+    val arrivalUrl = requestId(request.arrivalRequest.arrival.arrivalId)
     ArrivalMessageNotification(
+      s"$arrivalUrl/messages/${messageId.index}",
```
DavidGregory084, line 62: Should this really be `messageId.index`? I think that the message IDs that we expose in the URL are 1-indexed rather than 0-indexed. See e.g. https://github.com/hmrc/transit-movements-trader-at-destination/blob/master/app/models/MessageId.scala#L34 which is used for unbinding message IDs to URLs.

DavidGregory084: Perhaps we could refactor `MessageId` so that it has a method for getting the user-facing ID rather than an `unapply` method which is a bit magical IMO?

RobertLaverick: I'm adding a humanValue def that will return the index + 1 value, which is what I should be using here. I'll add some model tests to confirm this works since there are none for the notification.

crystalmark: Would it be easier to make the MessageId (etc) 1 based to save all the +1s and -1s and publicValue vs index?

DavidGregory084: The problem is that MessageId doesn't really exist - it's just the index of the message in the message array. There's no way of making it 1-based because it's a derived property. I am trying to fix that on CTDA-1374.

PR #246, opened 2021-08-20
CLI-71: Refactor for the push pull notification service and NCTScontroller

MatthewAlexanderBeck:


test/controllers/NCTSMessageControllerSpec.scala
```diff
@@ -42,275 +38,122 @@ import org.scalatestplus.scalacheck.ScalaCheckPropertyChecks
...
         val result = route(application, request).value
 
         status(result) mustEqual OK
-        header(LOCATION, result) mustBe Some(routes.MessagesController.getMessage(arrivalWithoutBox.arrivalId, arrivalWithoutBox.nextMessageId).url)
+        header(LOCATION, result) mustBe Some(routes.MessagesController.getMessage(arrival.arrivalId, arrival.nextMessageId).url)
       }
     }
 
-    "must return NotFound for an arrivalWithoutBox that does not exist" in {
+    "must return NotFound when the MovementMessageOrchestratorService returns an ArrivalNotFoundError" in {
```
ericohwotu, line 83: wasn't CLI-71 to return an OK in this scenario?

ericohwotu (approved):

app/services/PushPullNotificationService.scala
```diff
@@ -47,17 +54,30 @@ class PushPullNotificationService @Inject()(connector: PushPullNotificationConne
...
+            .map {
+              case Left(UpstreamErrorResponse(message, statusCode, _, _)) =>
+                logger.warn(s"Error $statusCode received while sending notification for boxId ${box.boxId}: $message")
+              case Right(_) => ()
+
+            }
+            .recover {
+              case NonFatal(e) =>
+                logger.error(s"Error while sending push notification", e)
+            }
```
DavidGregory084, line 80: It does feel a little bit like we are moving some HTTP-level logic (what I'd usually consider controller stuff) into the service layer here - not a comment that needs to be addressed but maybe something we need to think about

test/services/MovementMessageOrchestratorServiceSpec.scala
```diff
@@ -0,0 +1,140 @@
...
+        when(mockSaveMessageService.saveInboundMessage(any(), any())(any())).thenReturn(Future.successful(Right(())))
+        when(mockPushPullNotificationService.sendPushNotification(any(), any())(any())).thenReturn(Future.unit)
+
+        val application = baseApplicationBuilder
+          .overrides(bind[InboundRequestService].toInstance(mockInboundRequestService))
+          .overrides(bind[SaveMessageService].toInstance(mockSaveMessageService))
+          .overrides(bind[PushPullNotificationService].toInstance(mockPushPullNotificationService))
+          .build()
+
+        running(application) {
```
DavidGregory084, line 70: I think we should avoid running up a Guice apps for every test step as much as we can - I have seen this approach take test suites into tens of minutes just due to the overhead of starting apps. 
I don't think you need to run the app unless you want to hit a controller endpoint via the router - the rest of the time you can declare the application once and then use the `app.injector` to get things.

DavidGregory084 (approved):
Looks great, just a couple minor comments 💯

test/controllers/NCTSMessageControllerSpec.scala
amaalali: We've since pivoted and this 404 to 200 mapping is now done by the `transit-movements-trader-router`

app/services/PushPullNotificationService.scala
amaalali: All of the logic is the same from before, except instead of passing in the values boxId and notification, we're making this service dependent on needing a InboundMessageRequest which has those bits of data in it or has the data which we should work out in this service (eg using `ArrivalMessageNotification.fromInboundRequest`). 

Is that the main concern that we're passing in the headers to this service layer?

amaalali:
test this please

app/services/PushPullNotificationService.scala
DavidGregory084: Yeah I think so, but let's leave as is for now as like I say it's not a huge problem

PR #248, opened 2021-08-24
CTDA-1580 Upgrade auth-client for Auth CL250 changes

DavidGregory084:


app/config/Constants.scala
```diff
@@ -17,6 +17,7 @@
 package config
 
 object Constants {
-  val BoxName         = s"customs/transits##1.0##notificationUrl"
-  val XClientIdHeader = "X-Client-Id"
+  val BoxName          = s"customs/transits##1.0##notificationUrl"
```
jaynabonne, line 20: Nit-picky: does this the "s"?

app/connectors/MessageConnector.scala
```diff
@@ -30,56 +28,65 @@ import models.MessageSender
...
-    lazy val messageSender = MessageSender(arrivalId, message.messageCorrelationId)
+    val xmlMessage = TransitWrapper(message.message).toString
+
+    val messageSender = MessageSender(arrivalId, message.messageCorrelationId)
 
-    val newHeaders = headerCarrier
-      .copy(authorization = None)
-      .withExtraHeaders(addHeaders(message.messageType, dateTime, messageSender, channelType): _*)
+    val requestHeaders =
+      hc.headers(Seq(Constants.XClientIdHeader, Constants.XRequestIdHeader)) ++ messageHeaders(message.messageType, dateTime, messageSender, channelType)
```
jaynabonne, line 57: So if I read this right, it only needs two headers from the original hc?

app/config/Constants.scala
DavidGregory084: No you are right - I think this was left over from when we considered parsing the API version header - will update

app/connectors/MessageConnector.scala
DavidGregory084: Yes I think so - the eis-router adds its own Authorization header because there is a specific bearer token required for each EIS instance. The other one that we pass around everywhere is for determining which channel the movement came from (the frontend services or the API), which is not so relevant for the router.

PR #252, opened 2021-09-07
CTCTRADERS-2620 case insenitive

shaileshp0110:


jaynabonne (approved):

app/repositories/ArrivalMovementRepository.scala
```diff
@@ -309,7 +309,7 @@ class ArrivalMovementRepository @Inject()(
     }
 
   private def withMRNSearchQuery(mrn: String, pageSize: Option[Int], channelFilter: ChannelType, selector: JsObject, countSelector: JsObject) = {
-    val mrnSelector = Json.obj("movementReferenceNumber" -> Json.obj("$regex" -> mrn))
+    val mrnSelector = Json.obj("movementReferenceNumber" -> Json.obj("$regex" -> mrn, "$options" -> "i"))
```
DavidGregory084, line 312:
```suggestion
    val mrnSelector = Json.obj("movementReferenceNumber" -> Json.obj("$regex" -> s"^$mrn", "$options" -> "i"))
```
This might help the performance by ensuring that the regex search is always a prefix search - unless you want inner matches?

jaynabonne: I believe the match is supposed to be anywhere in the string.

DavidGregory084 (approved):

app/repositories/ArrivalMovementRepository.scala
shaileshp0110: I am not sure if the MRN  can be always prefixed search it's a good point, @amaalali  do you it free search or it will be prefixed search

PR #259, opened 2021-09-29
CTCTRADERS-2607 Check new enrolment and legacy enrolment

carrielies:


DavidGregory084:
@carrielies there is already a PR for this change - see #257

DavidGregory084:
we have also covered departures at https://github.com/hmrc/transits-movements-trader-at-departure/pull/138 and the test support API at https://github.com/hmrc/common-transit-convention-traders-test-support/pull/78/files

PR #260, opened 2021-10-05
CTCTRADERS-2636: Removed message status from arrival response

MatthewAlexanderBeck:


app/models/ArrivalWithoutMessages.scala
```diff
@@ -36,7 +36,8 @@ case class ArrivalWithoutMessages(
   lastUpdated: LocalDateTime = LocalDateTime.now,
   notificationBox: Option[Box],
   nextMessageId: MessageId,
-  nextMessageCorrelationId: Int
+  nextMessageCorrelationId: Int,
+  messagesMetaData: Seq[MessageMetaData]
```
jaynabonne, line 40: Nit pick: it seems strange to still call this ArrivalWithoutMessages. :)

MatthewAlexanderBeck: Yeah, it looks a bit strange however it doesnt return the full message, only the message type and date. The other corresponding model will return the full message with all data. Saves us having to get too much when creating summary pages etc that dont require all the data

jaynabonne: Seems fine.

jaynabonne (approved):

PR #271, opened 2021-12-16
CTDA-1585: Pull only required messages for PDFGenerationController

drnaylor-oc:
* Adds `getMessagesOfType` to `ArrivalMovementRepository` for retriving only messages and an EORI number
* Adds `AuthenticatedGetMessagesAction` and associated support for handling requests for specific message types
* Adds `ArrivalMessages` object to bundle messages, the EORI number and arrival ID
* Adds supporting tests

A similar PR will be made to the departure repo.

### DIFF
--- a/app/services/UnloadingPermissionPDFService.scala
+++ b/app/services/UnloadingPermissionPDFService.scala
@@ -18,6 +18,8 @@ package services
 
 import connectors.ManageDocumentsConnector
 import models.Arrival
+import models.MovementMessage
+import models.PdfDocument
 import models.WSError
 import models.WSError._
 import play.api.mvc.Results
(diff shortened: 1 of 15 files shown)


RobertLaverick (approved):

app/services/UnloadingPermissionPDFService.scala
```diff
@@ -28,33 +30,38 @@ import uk.gov.hmrc.http.HeaderCarrier
...
-                Left(OtherError(otherErrorCode, result.body))
-            }
-        }
-      case None => Future.successful(Left(NotFoundError))
+      case Some(unloadingPermission) => generatePDF(unloadingPermission.message)
+      case None                      => Future.successful(Left(NotFoundError))
     }
 
+  def getPDF(messages: List[MovementMessage])(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[Either[WSError, PdfDocument]] =
+    generatePDF(messages.minBy(_.messageCorrelationId).message)
```
DavidGregory084, line 65: I think we would usually get the `.maxBy(_.messageCorrelationId)` i.e. the most recent message?

DavidGregory084 (approved):
This is looking really good @drnaylor-oc - just one small query 👍

app/services/UnloadingPermissionPDFService.scala
drnaylor-oc: I took this from https://github.com/hmrc/transit-movements-trader-at-destination/blob/main/app/services/ArrivalMessageSummaryService.scala#L102, which suggests that we get the first message rather than the last. Is it possible that this in itself is a bug? I'd imagine it probably is because everything else is max rather than min.

Might be good for me to introduce another test in here to back up what the right behaviour is.

DavidGregory084: Resolving as this seems to be deliberate https://jira.tools.tax.service.gov.uk/browse/CTDA-1144

PR #278, opened 2022-01-18
CTCTRADERS-2895: Added index for fetching all arrivals with a date AN…

jackheslop96:
…D movement reference number filter.


DavidGregory084 (approved):
👍

PR #279, opened 2022-01-19
CTCTRADERS-2872: remove message metadata from  get Arrivals

kenneth-rayner:


MatthewAlexanderBeck (approved):

drnaylor-oc (approved):
The removal of this field shouldn't impact the API frontend as we don't even look for it - so this looks good to me.

PR #281, opened 2022-01-25
Log full stacktraces when there are errors inserting into Mongo

DavidGregory084:


drnaylor-oc (approved):
I don't understand why the status check test failed (local tests all return as passes) - but regardless of that this looks fine.

DavidGregory084:
rebuild

DavidGregory084:
Unfortunately it seems like we have a flaky test

PR #284, opened 2022-01-31
CTDA-1669: Add metrics in the PPNS connector, fraction of posted messages that have boxes attached

drnaylor:
Thoughts:

* ~~I suspect what we're interested in is the ratio of calls to getBox vs sendPushNotification over the entire lifetime of the application. In this case, a ratio of two counters is what we want. However, if we want it over a rolling time period, we might be able to do this with Meters.~~
* ~~Do we just want a ratio of calls to these methods, or do we want a ratio of _successful_ calls? My interpretation is the former, but if it's the latter, should be easy to do.~~

Departures will follow once we're happy with this.

---

This has been through a lot of iterations, we settled on the fraction of submitted arrivals that have a box attached. We also have a week sliding window for this data.

### DIFF
--- a/test/audit/AuditServiceSpec.scala
+++ b/test/audit/AuditServiceSpec.scala
@@ -19,6 +19,7 @@ package audit
 import base.SpecBase
 import cats.data.Ior
 import config.Constants
+import generators.ModelGenerators
 import models.ChannelType.api
 import models._
 import models.request.AuthenticatedRequest
@@ -40,7 +41,7 @@ import utils.XMLTransformer.toJson
 
 import java.time.LocalDateTime
 
-class AuditServiceSpec extends SpecBase with ScalaCheckPropertyChecks with BeforeAndAfterEach {
+class AuditServiceSpec extends SpecBase with ScalaCheckPropertyChecks with BeforeAndAfterEach with ModelGenerators {
 
   val mockAuditConnector: AuditConnector = mock[AuditConnector]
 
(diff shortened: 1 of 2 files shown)


RobertLaverick (approved):

drnaylor-oc:
The two commits in this PR at this stage should be considered as two separate solutions.

RobertLaverick (approved):

RobertLaverick:
test me again

RobertLaverick:
retest this please

drnaylor-oc:
retest

app/audit/AuditDetails.scala
```diff
@@ -18,20 +18,30 @@ package audit
...
+}
+
 sealed abstract class AuditDetails {
   def channel: ChannelType
   def customerId: String
   def json: JsObject
 }
 
-case class AuthenticatedAuditDetails(channel: ChannelType, enrolmentId: Ior[TURN, EORINumber], json: JsObject) extends AuditDetails {
+case class AuthenticatedAuditDetails(channel: ChannelType, enrolmentId: Ior[TURN, EORINumber], json: JsObject, boxId: Option[BoxId]) extends AuditDetails {
```
RobertLaverick, line 44: have you considered having boxId default to None?
`boxId: Option[BoxId] = None`

RobertLaverick (changes requested):

RobertLaverick (approved):

PR #292, opened 2022-03-25
CTDA-1886: Clean up audit models, pull logic out to the audit service

drnaylor-oc:
Awaiting CTDA-1885 so the change from there can be incorporated.


app/audit/AuditDetails.scala
```diff
@@ -16,56 +16,58 @@
...
+  )(unlift(AuthenticatedAuditDetails.unapply))
 }
 
 case class UnauthenticatedAuditDetails(channel: ChannelType, customerId: String, json: JsObject) extends AuditDetails
 
 object UnauthenticatedAuditDetails {
 
-  implicit val writes: OWrites[UnauthenticatedAuditDetails] = _.writeAuditDetails
+  implicit val writes: OWrites[UnauthenticatedAuditDetails] = (
+    (__ \ "channel").write[ChannelType] and
```
liandra-giacomine, line 50: We seem to have a few repeated strings in the code, is it worth adding a file for constants if we don't already have one?

liandra-giacomine (approved):

app/audit/AuditDetails.scala
drnaylor-oc: More my opinion here, but in this scenario, no. For example - if this "channel" changed here, it is only because this specific model has had a field name change. Thus, adding it as a constant will add no value as there is actually no link between all the places we specify it - and if we change one field name, while we might change all the models with the same field name, we aren't necessarily going to do that.

If the string was a value that, if it changed, it's likely it should be changed in all other places its specified, then it should be a constant - that's not the case here.

Additionally here, it's far easier to see the link between the case class field name and the string.

PR #296, opened 2022-06-14
CTDA-1935: Drop and recreate the TTL index whenever the TTL is updated

drnaylor-oc:
I've looked at adding a test for this, but due to IntelliJ not being able to cope with reactive mongo, and reactive mongo's judicious use of structural typing, writing tests for this might be a little convoluted. 

I have tested this locally, when the index TTL is wrong it does get dropped and recreated. When it's correct, I get the log message telling me otherwise.

I'm going to see if I can figure out a reasonable way to test this in isolation but it may not be that easy to do as we'll need to make sure we test all scenarios, and with the way the class is set up, that's not easy to do.

### DIFF
--- a/app/repositories/ArrivalMovementRepository.scala
+++ b/app/repositories/ArrivalMovementRepository.scala
@@ -560,7 +560,7 @@ class ArrivalMovementRepository @Inject()(
         val indexToDrop = indexes
           .filter(_.name.contains(lastUpdatedIndexName))
           .filter(x => x.expireAfterSeconds.map(_ != appConfig.cacheTtl).getOrElse(false))
-          .get(0)
+          .headOption
         indexToDrop match {
           case Some(index) =>
             val ttl = index.expireAfterSeconds.map(x => s"$x seconds").getOrElse("unset")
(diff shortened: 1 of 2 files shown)


app/repositories/ArrivalMovementRepository.scala
```diff
@@ -557,18 +557,23 @@ class ArrivalMovementRepository @Inject()(
...
-            .map(
-              _ => true
-            )
-        } else {
-          logger.info(s"$oldLastUpdatedIndexName does not exist or has already been dropped")
-          Future.successful(true)
+        val indexToDrop = indexes
+          .filter(_.name.contains(lastUpdatedIndexName))
+          .filter(x => x.expireAfterSeconds.map(_ != appConfig.cacheTtl).getOrElse(false))
+          .get(0)
```
RobertLaverick, line 563: why get(0) rather than just .head?

RobertLaverick: or headOption

drnaylor-oc: Yeah, me being daft. I'll switch it to headOption

RobertLaverick (approved):

PR #300, opened 2022-12-08
CTDA-1973: Added received parameter to message object

liandra-giacomine:


app/models/MovementMessage.scala
```diff
@@ -87,11 +90,12 @@ object MovementMessageWithStatus extends NodeSeqFormat with MongoDateTimeFormats
     (
       (__ \ "messageId").read[MessageId] and
         (__ \ "dateTime").read[LocalDateTime] and
+        (__ \ "received").read[LocalDateTime] and
```
drnaylor-oc, line 93: I've just realised there is a significant issue with this -- if a movement does not have a received date/time then the reads on this will fail. We're not going back to patch in the received time, so we _cannot_ merge this as is.

drnaylor-oc (changes requested):
This can't be merged like this as this will break production.

drnaylor-oc (dismissed):

drnaylor-oc (approved):

